PR #7, opened 2017-01-16
Adds panda

marialivia16:
@susiecoleman

### DIFF
--- a/app/controllers/PanDomainAuthActions.scala
+++ b/app/controllers/PanDomainAuthActions.scala
@@ -14,7 +14,7 @@ trait PanDomainAuthActions extends AuthActions {
 
   override def domain: String = Config.pandaDomain
 
-  override lazy val system: String = "atom-workshop"
+  override lazy val system: String = Config.pandaSystem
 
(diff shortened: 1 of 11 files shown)


app/controllers/PanDomainAuthActions.scala
```diff
@@ -0,0 +1,21 @@
...
+trait PanDomainAuthActions extends AuthActions {
+
+  override def validateUser(authedUser: AuthenticatedUser): Boolean =
+    (authedUser.user.email endsWith "@guardian.co.uk") && authedUser.multiFactor
+
+  override def authCallbackUrl: String = Config.pandaAuthCallback
+
+  override def domain: String = Config.pandaDomain
+
+  override lazy val system: String = "atom-workshop"
```
philmcmahon: Could this be imported from config rather than hard coded?

philmcmahon:
Other than the above comment, 👍

app/controllers/PanDomainAuthActions.scala
marialivia16: Sounds reasonable. I'll add it to the config.

PR #9, opened 2017-01-17
[Do not merge] Add template to create dynamo tables, link with main template

philmcmahon:
Andy has now cloudformed 3 stacks for the DEV, CODE and PROD dynamo tables. The main stack then infers the name of the tables based off the db stack name being passed as a parameter.

Queries to the tables will require an atomType and an id.


philmcmahon:
Don't merge this as its non-generic and I haven't saved the generic stuff into a different repo yet!

PR #10, opened 2017-01-17
FE build pipeline

clloyd:
@philmcmahon 

This adds a `npm run start` and `npm run build` command.

TODO:
- [x] Readme


yarn.lock
```diff
@@ -0,0 +1,3271 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
ShaunYearStrong: Doesn't this file need ignoring?

setup.sh
```diff
@@ -0,0 +1,24 @@
...
+    echo -e "Yarn is not required for the application (but is for this script)"
+    echo -e "Packages can be manually installed with npm\n\r\n\r"
+
+    exit 1
+fi
+
+printf "\n\rSetting up client side dependancies... \n\r\n\r"
+printf "\n\rInstalling NPM packages via yarn... \n\r\n\r"
+
+yarn
```
ShaunYearStrong: 🎉

ShaunYearStrong:
👍

yarn.lock
clloyd: Nope! One of the advantages of yarn is the lock file ensures we all use the exact same dependencies, designed to be checked in https://yarnpkg.com/en/docs/yarn-lock

PR #11, opened 2017-01-18
Remove DEV specific properties from application.conf

philmcmahon:
This is a tiny change. People will need to run ./fetch-config.sh to get the app working again after pulling this change.


clloyd:
merge in master to get build to pass

PR #14, opened 2017-01-19
Jest testing implementation

clloyd:
Very curious to get feedback on this, I've added a general utility component (Modal) and a basic test suite.

This is making use of `jest` and it's snapshotting ability to ensure that the modal renders correctly when open and closed, it's separately making use of `enzyme` to shallow render the component and simulate the click events.

FYI - The `__snapshots__` folder is autogenerated when you run the tests

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,5 @@
+// .babelrc
+{
+  "presets": ["es2015", "react"],
+  "plugins": ["transform-object-assign"]
+}
--- a/conf/webpack.conf.js
+++ b/conf/webpack.conf.js
@@ -8,7 +8,7 @@ module.exports = {
             {
                 test:    /\.js$/,
                 exclude: /node_modules/,
-                loaders: ['babel-loader?presets[]=es2015&presets[]=react&plugins[]=transform-object-assign']
+                loaders: ['babel-loader']
             },
             {
                 test: /\.scss$/,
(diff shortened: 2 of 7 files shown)


ShaunYearStrong:
> FYI - The __snapshots__ folder is autogenerated when you run the tests

Does this mean the snapshots themselves are autogenerated from the component too?

clloyd:
@ShaunYearStrong Yeah exactly, when you first run the test it generates the snapshots... then subsequents runs compares, if they are wrong the tests fail... but you can regenerate the snapshots with `yarn -n`, the idea is to save the whole `.contains("<div>Test Content</div>")`. And you can "approve" the snapshots in the PR.

ShaunYearStrong:
@clloyd I get you, it seems odd but I think it _could_ be great. i guess after this goes in, we'll set them up in CI so if someone's forgotten to generate they'll fail in here?

clloyd:
I think having CI fail if there's un-generated snapshots

ShaunYearStrong:
Yeah thats what I meant haha, I'm just used to getting CI updates from inside github. I think the additional step could become a pain but there's always Githooks!

Aside from that, I like it, let's see how it goes 👍

clloyd:
I've merged this... but it's easy to remove if it ends up being a bad idea

PR #17, opened 2017-01-24
"React/Reduxify" the base page

ShaunYearStrong:
Added React routing, reduction, a store provider etc to the project, as I think it's pretty obvious we're going with React/Redux.

picture 74


public/js/util/storeAccessor.js
```diff
@@ -0,0 +1,9 @@
+let savedStore;
+
+export function setStore(store) {
```
ShaunYearStrong: Is there a way of getting rid of this yet? /cc @clloyd

clloyd: We only use it to get config out of the store in non-react utils. If we can do that in an alternative way... then yes!

PR #25, opened 2017-01-25
Some basic form components

ShaunYearStrong:
Kind of jumping the gun and just making some generic stuff.

picture 80

### DIFF
--- a/public/js/components/FormFields/FormFieldRadioButtons.js
+++ b/public/js/components/FormFields/FormFieldRadioButtons.js
@@ -1,0 +1,34 @@
+import React from 'react';
+
+export default class FormFieldRadioButtons extends React.Component {
+
+  renderButton(value, i) {
+    return (
+      <div key={i} className="form__group form__group--radio">
+        <input className="form__radio-btn" type="radio" name={this.props.fieldName} value={value} defaultChecked={this.props.checkedValue === value ? 'checked' : false} />
+        <span className="form__label form__label--radio">{value}</span>
+      </div>
+    )
+  }
+
+  renderButtons() {
+    return this.props.fieldValues.map(this.renderButton, this);
+  }
+
+  render() {
+    return (
+        <div>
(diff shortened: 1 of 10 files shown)


public/js/components/FormFields/FormFieldRadioButtons.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+
+export default class FormFieldRadioButtons extends React.Component {
+
+  renderButton(value, i) {
+    return (
+      <div key={i} className="form__group form__group--radio">
+        <input className="form__radio-btn" type="radio" name={this.props.fieldName} value={value} defaultChecked={this.props.checkedValue === value ? 'checked' : false} />
```
clloyd: should this be `checked="true"` (rather than defaultChecked)

public/js/components/FormFields/FormFieldRadioButtons.js
```diff
@@ -0,0 +1,34 @@
...
+    return (
+        <div>
+          <label className="form__label" htmlFor={this.props.fieldName}>{this.props.fieldLabel}</label>
+          {this.renderButtons()}
+        </div>
+    )
+  }
+}
+
+FormFieldRadioButtons.propTypes = {
```
clloyd: With property Initialiser plugin this can be `static propTypes = {` within the class... up to you though!

public/js/components/FormFields/FormFieldRadioButtons.js
ShaunYearStrong: There was a warning that `checked` without an `onClicked` handler would lead to the radio buttons being unable to change.

public/js/components/FormFields/FormFieldRadioButtons.js
ShaunYearStrong: Makes sense!

ShaunYearStrong:
@clloyd - Only just saw your part about the tests. I will do!

PR #26, opened 2017-01-31
Create scribe component

ShaunYearStrong:
picture 83

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,12 @@
     "react-redux": "^5.0.2",
     "react-router": "^3.0.2",
     "redux": "^3.6.0",
-    "redux-thunk": "^2.2.0"
+    "redux-thunk": "^2.2.0",
+    "scribe": "guardian/scribe#fcf4e16",
+    "scribe-plugin-link-prompt-command": "^1.0.0",
+    "scribe-plugin-noting": "^0.3.28",
+    "scribe-plugin-sanitizer": "^0.1.10",
+    "scribe-plugin-toolbar": "^1.0.0",
+    "scribe-plugin-keyboard-shortcuts": "^0.1.1"
   }
 }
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -44,6 +44,12 @@
     "react-redux": "^5.0.2",
     "react-router": "^3.0.2",
     "redux": "^3.6.0",
-    "redux-thunk": "^2.2.0"
+    "redux-thunk": "^2.2.0",
+    "scribe": "guardian/scribe#fcf4e16",
```
ShaunYearStrong: This will need changing when there's a new release of scribe

public/js/components/ReactApp.js
```diff
@@ -1,41 +1,37 @@
...
-import FormFieldTextInput from './FormFields/FormFieldTextInput';
-import FormFieldSelectBox from './FormFields/FormFieldSelectBox';
-import FormFieldRadioButtons from './FormFields/FormFieldRadioButtons';
+import ScribeEditor from './ScribeEditor/ScribeEditor';
 
 export default class ReactApp extends React.Component {
 
-  updateCheckbox = () => {
-    console.log("Updated");
+  updateDescription = (newContent) => {
```
ShaunYearStrong: just an example.

ShaunYearStrong:
I'm going to remove noting for now. The broken build is caused by it having a dependency on bower.

clloyd:
Is this missing styles?

ShaunYearStrong:
@clloyd Kind of, it doesn't have any buuuut I didn't want to make assumptions about the usage of scribe in here. I think we're coming to a point where we're going to have to decide on some kind of page structure here

clloyd:
The general component doesnt need to have styles, but we should pass in a relevant class and style that in the example

PR #27, opened 2017-01-31
Add GET, POST endpoints for atoms

philmcmahon:
PUT still to come. Using the lovely API design suggested by @adamnfish. 

As part of this I had to write rather a lot of boilerplate in AtomDataStores.scala, hopefully we can kill this off at some point in the future - its about 6 lines per type of atom so not terrible but not ideal.

(@ShaunYearStrong - as we have to use Circe for nice Atom to Json serialising I thought I'd be consistent and set up to use that everywhere, and so ended up removing the play json lib you added. Let me know if this breaks something and I'll put it back!)

### DIFF
--- a/app/db/AtomWorkshopDB.scala
+++ b/app/db/AtomWorkshopDB.scala
@@ -1,13 +1,19 @@
 package db
 
 import com.amazonaws.services.dynamodbv2.model.AmazonDynamoDBException
-import com.gu.atom.data.{DynamoCompositeKey, DynamoDataStore}
+import com.gu.atom.data.{DataStoreError, DataStoreResult, DynamoCompositeKey, DynamoDataStore}
 import com.gu.contentatom.thrift.{Atom, AtomData, AtomType, ContentChangeDetails}
 import com.gu.contentatom.thrift.atom.cta.CTAAtom
 import com.gu.contentatom.thrift.atom.explainer.{DisplayType, ExplainerAtom}
 import com.gu.contentatom.thrift.atom.media.MediaAtom
 import play.api.Logger
 import cats.syntax.either._
+import com.gu.atom
+import models.{AtomAPIError, AtomWorkshopDynamoDatastoreError, CreateAtomDynamoError}
+import com.gu.fezziwig.CirceScroogeMacros._
+import io.circe._
+import io.circe.syntax._
+
 
 object AtomWorkshopDB {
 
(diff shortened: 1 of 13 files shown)


app/db/AtomWorkshopDB.scala
```diff
@@ -0,0 +1,48 @@
...
+      Logger.info(s"Successfully created atom of type ${atomType.name} with id ${defaultAtom.id}")
+      r
+    } catch {
+      case e: AmazonDynamoDBException =>
+        Logger.error("Atom creation failed: ", e)
+        Left(e)
+    }
+  }
+
+  def getAtom(datastore: DynamoDataStore[_ >: ExplainerAtom with CTAAtom with MediaAtom], atomType: AtomType, id: String) = {
```
adamnfish: Is this function useful?

conf/routes
```diff
@@ -6,4 +6,10 @@ GET     /oauthCallback              controllers.Login.oauthCallback
...
 
 # Map static resources from the /public folder to the /assets URL path
-GET     /assets/*file               controllers.Assets.versioned(path="/public", file: Asset)
\ No newline at end of file
+GET     /assets/*file               controllers.Assets.versioned(path="/public", file: Asset)
+
+
+GET  /api/:atomType/:id/preview             controllers.App.getAtom(atomType: String, id: String, version = "preview")
+
+# We only ever create into the preview table so no version needed here
```
adamnfish: is it worth requiring preview in the URL for consistency anyway?

conf/routes
```diff
@@ -6,4 +6,10 @@ GET     /oauthCallback              controllers.Login.oauthCallback
 GET     /reauth                     controllers.Login.reauth
 
 # Map static resources from the /public folder to the /assets URL path
-GET     /assets/*file               controllers.Assets.versioned(path="/public", file: Asset)
\ No newline at end of file
+GET     /assets/*file               controllers.Assets.versioned(path="/public", file: Asset)
+
+
+GET  /api/:atomType/:id/preview             controllers.App.getAtom(atomType: String, id: String, version = "preview")
```
adamnfish: This isn't a hugely restful choice of URL, but no worries if it works nicely.

app/models/APIResponse.scala
```diff
@@ -0,0 +1,31 @@
...
+
+object APIResponse extends Results {
+  def exceptionToResult(e: Exception) = {
+    Logger.error(e.getMessage, e)
+    InternalServerError(AtomWorkshopAPIResponse(e.getMessage).asJson.noSpaces)
+  }
+
+  def apply[T](result: Either[Exception, T])(implicit encoder: Encoder[T]): Result = {
+    val res = result.fold({
+      case e: DataStoreError => exceptionToResult(e)
```
adamnfish: Since you handle all these the same why, why do this match? Surely the following would suffice?

    exceptionToResult(_)

ShaunYearStrong:
@philmcmahon - Should all be good, only needed it for serialising!

app/db/AtomWorkshopDB.scala
```diff
@@ -0,0 +1,48 @@
...
+import cats.syntax.either._
+
+object AtomWorkshopDB {
+
+  def buildKey(atomType: AtomType, id: String) = DynamoCompositeKey(atomType.name, Some(id))
+
+  def buildDefaultAtom(atomType: AtomType) = {
+    val defaultAtoms: Map[AtomType, AtomData] = Map(
+      AtomType.Explainer -> AtomData.Explainer(ExplainerAtom("-", "-", DisplayType.Flat)),
+      AtomType.Cta -> AtomData.Cta(CTAAtom(""))
```
susiecoleman: Should there be one for media atom here?

app/db/AtomWorkshopDB.scala
philmcmahon: No. I'll kill it!

app/models/APIResponse.scala
philmcmahon: Good point.

app/db/AtomWorkshopDB.scala
philmcmahon: No because eventually we won't support media atoms I think - I included them in the other bit because currently the only test atom we have is a media one.

philmcmahon:
Tested this on CODE. it works!

PR #29, opened 2017-02-02
Add tests for form fields

ShaunYearStrong:


public/js/components/FormFields/__snapshots__/FormFieldRadioButtons.spec.js.snap
```diff
@@ -0,0 +1,37 @@
...
+      One
+    </span>
+  </div>
+  <div
+    className="form__group form__group--radio">
+    <input
+      checked={false}
+      className="form__radio-btn"
+      name="test"
+      onChange={undefined}
```
clloyd: Do we want to pass an update function so we can verify it's being used in correct place?

PR #32, opened 2017-02-03
Add update route

philmcmahon:
The unexciting way to update atoms. Maria's single-field update to follow! This route takes a blob of json looking like this: (note the 'cta' block inside the 'data' block)

```
{"id":"dde9f8bc-f5cb-4e8b-a8ba-0914a15b8adc",
"atomType":"cta",
"labels":["asdasdasda"],
"defaultHtml":"<div></div>",
"data":{
    "cta":{"url":"asdadsas"}
},
"contentChangeDetails":{"revision":1}}
```

And, if an atom with that id and type exists, updates it. Big thanks to @tomrf1 for getting fezziwig to work with scanamo's version of cats!

### DIFF
--- a/app/controllers/App.scala
+++ b/app/controllers/App.scala
@@ -4,12 +4,12 @@ import config.Config
 import models._
 import play.api.Logger
 import play.api.libs.ws.WSClient
-import play.api.mvc._
+import play.api.mvc.Controller
 import cats.syntax.either._
+import com.gu.contentatom.thrift.Atom
 import db.{AtomDataStores, AtomWorkshopDB}
 import com.gu.fezziwig.CirceScroogeMacros._
 import io.circe._
-import io.circe.generic.auto._
 import io.circe.syntax._
 import util.HelperFunctions._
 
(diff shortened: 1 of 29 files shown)


app/util/HelperFunctions.scala
```diff
@@ -14,4 +23,31 @@ object HelperFunctions {
...
+    Logger.error(atomApiError.msg, exception)
+    Left(atomApiError)
+  }
+
+  def extractRequestBody(body: Option[String]): Either[AtomAPIError, String]= {
+    Either.cond(body.isDefined, body.get, BodyRequiredForUpdateError)
+  }
+
+  def parseAtomJson(atomJson: String): Either[AtomAPIError, Atom] = {
+    println(atomJson)
```
philmcmahon: bad phil!

susiecoleman: What Phil said

app/controllers/App.scala
```diff
@@ -35,14 +34,27 @@ class App(val wsClient: WSClient) extends Controller with PanDomainAuthActions {
...
 
+  def updateEntireAtom(atomType: String, id: String) = AuthAction { req =>
+    APIResponse {
+      for {
+        atomType <- validateAtomType(atomType)
+        payload <- extractRequestBody(req.body.asText)
+        newAtom <- parseAtomJson(payload)
+        datastore <- AtomDataStores.getDataStore(atomType, Preview)
+        currentAtom <- AtomWorkshopDB.getAtom(datastore, atomType, id)
+        result <- AtomWorkshopDB.updateAtom(datastore, atomType, req.user, currentAtom,newAtom)
```
susiecoleman: I wonder if we should be returning the atom in the case of create and update? I realise this would require a change to the atom-maker library so I don't think it should be blocker to getting this out I'm just curious why the decision was made to return unit...

PR #33, opened 2017-02-03
Grid Component

clloyd:
screen shot 2017-02-03 at 15 25 00

This adds a grid component complete with tests. 

Also adds a `./scripts/start.sh` startup for both play and fe asset compilation


public/js/components/FormFields/FormFieldImageSelect.js
```diff
@@ -0,0 +1,123 @@
...
+    window.removeEventListener('message', this.onMessage, false);
+  }
+
+  openModal = () => {
+      this.setState({ modalOpen: true });
+      window.addEventListener('message', this.onMessage, false);
+  }
+
+  validMessage(data) {
+    return data && data.crop && data.crop.data && data.image && data.image.data;
```
ShaunYearStrong: I hate JS sometimes

public/js/components/FormFields/FormFieldImageSelect.js
```diff
@@ -0,0 +1,123 @@
...
+          </button>
+        </div>
+      </div>
+    )
+  }
+
+  render () {
+    return (
+      <div>
+        <label className="form__label" htmlFor={this.props.fieldName}>{this.props.fieldLabel}</label>
```
ShaunYearStrong: I've just realised I didn't do this on mine, but the value of `htmlFor` should be the same as the `id` of the element we're referring to.

public/js/util/imageHelpers.js
```diff
@@ -0,0 +1,56 @@
...
+export function parseImageFromGridCrop(cropData) {
+  return {
+    assets: cropData.assets.map(parseAsset),
+    master: parseAsset(cropData.master),
+    mediaId: cropData.specification.uri
+  }
+}
+
+export function findSmallestAsset(assetsArray) {
+  return assetsArray.reduce((smallestAsset, newAsset) => {
```
ShaunYearStrong: Couldn't we use `Array.prototype.sort()` instead? https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

ShaunYearStrong: Ignore that, looks like this way is faster https://jsperf.com/reduce-vs-sort

clloyd: Yeah reduce has a complexity of `n` vs sort which can be a lot more

public/js/components/FormFields/FormFieldImageSelect.js
clloyd: Fixing

PR #34, opened 2017-02-03
Switch from Reqwest and Q to Fetch and Promises

ShaunYearStrong:
pandaReqwest is dead. Long live pandaFetch!!!


package.json
```diff
@@ -39,19 +39,21 @@
...
-    "q": "^1.4.1",
     "react": "^15.4.2",
     "react-dom": "^15.4.2",
     "react-redux": "^5.0.2",
     "react-router": "^3.0.2",
     "redux": "^3.6.0",
     "redux-thunk": "^2.2.0",
-    "reqwest": "^2.0.5",
     "scribe": "guardian/scribe#fcf4e16",
+    "scribe-plugin-keyboard-shortcuts": "^0.1.1",
```
ShaunYearStrong: Yarn likes alphabetical order...

public/js/services/AtomsApi.js
```diff
@@ -1,21 +1,29 @@
...
 
 export default {
 
   fetchAtom: (atomId, atomType) => {
-    return pandaReqwest({
-      url: `/api/preview/${atomType}/${atomId}`,
-      method: 'get'
-    });
+    return pandaFetch(
+      new Request(
```
ShaunYearStrong: https://developer.mozilla.org/en-US/docs/Web/API/Request/Request

public/js/services/pandaFetch.js
```diff
@@ -0,0 +1,39 @@
+import { reEstablishSession } from 'panda-session';
+import { getStore } from '../util/storeAccessor';
+
+function checkStatus(res) {
```
ShaunYearStrong: This exists because unless there's a network error, fetch will resolve even with an error code.

ShaunYearStrong:
@clloyd - The build seems to be breaking because it's trying to test the image selector when it's not there. I think there's a problem with the build in CI.

PR #35, opened 2017-02-06
Generic update endpoint using Shapeless

marialivia16:
## Problem trying to solve
Given a generic route `/api/:atomType/:id/:path`, a JSON body and an object of type `A`, we want to update the field of the object found by accessing the specified path with the JSON value.
e.g.
Route: `/api/media/8ba4c6c8-0815-45d9-9e87-0a6c20954094/metadata.channelId`
Body: `{ "value": "some-awsome-new-id" }`
Object:
```scala
val atom = Media(MediaAtom(
     assets = List(
       Asset(
         assetType = Video,
         id = "https://www.gu.com/something", 
         platform = Url, 
         mimeType = None), 
       Asset(
         assetType = Video,
         id = "https://www.gu.com/something/else",
         platform = Url,
         mimeType = None)),
     title = "Amazing title",
     category = Documentary,
     duration = Some(9),
     metadata = Some(Metadata(
       tags = Some(List()), 
       commentsEnabled = Some(false), 
       channelId = Some("some-awesome-id")))))
```
The result is a new object that has the `channelId` field from inside `metadata` updated to the new value `some-awsome-new-id`.

## Why is this challenging?
This is a generic endpoint. We don't know the type of the object we want to update.
We have tried Scala reflection and Shapeless, hoping we could use a macro and generate some code to help us traverse the object. But in both cases we realised we don't have enough information about the types.
e.g.
Let's have a look at a snippet from the Shapeless attempt:
```scala
val atomLabelledGeneric = LabelledGeneric[Atom]
val genericRepr = atomLabelledGeneric.to(atom) 
val updatedGeneric = genericRepr + (Symbol("title") ->> "new title"))
```
Shapeless works on case classes and sealed traits. `Generic` represents the ability to convert from a concrete type to a generic `HList` representation of the type. `LabelledGeneric` is very similar to `Generic` but includes information about the fields to the raw structure. We want to use this because we need to know the name of the label in order to update it.

What happens next is we convert the `atom` object to a generic representation that looks something like this: 
```scala
"Amazing title" :: Documentary :: Some(9) :: Some(Metadata(Some(List()), Some(false), Some("some-awesome-id")))
```
Notice how this only converts the first level.
The last line in the example updates the field with the label `title` with the new value.
There are 2 issues with this implementation:
1. The `Symbol` constructor doesn't allow parameters - all the magic happens at compile time.
2. Because Shapeless only converts the first level we would have to recursively convert every level and without knowing the type of the next level we wouldn't be able to instantiate our `LabelledGeneric`.
Given all this we were unable to find an implementation that could be auto-generated.

## How did we make this work?
After some googling, we tried a new approach, one that uses Shapeless as well.
Given an object of type `A`, we:
- convert it to a structure composed of nested Scala maps;
- use the `path` from the route to navigate to the field that needs to be updated;
- do the update;
- convert the maps back to a class.

e.g.
We defined a type class `ToMapRec` that will convert the case class to a map recursively. 
This is an HList at a random point during our iteration: `"a" :: Some("b") :: 3 :: Seq("d", "e") :: hNil`
At the moment this type class has 5 cases (the last 2 cases are more specific):
1. One for heads that don't have ToMapRec instances
2. One for `hNil` that just returns an empty map
3. One for when we know how to convert the tail of the record, and we know that the head is something that we can also recursively convert.
4. One for when the head is an `Option`
5. One for when the head is a `Seq`

Let's consider the `atom` we declared at the beginning. This is how the map will look like: 
```scala
val map = Map(
  "duration" -> Some(9),  
  "source" -> None, 
  "assets" -> List(
    Map(
      "assetType" -> Video, 
      "_passthroughFields" -> Map(), 
      "mimeType" -> None, 
      "version" -> 8, 
      "id" -> "https://www.gu.com/something", 
      "platform" -> Url), 
    Map(
      "assetType" -> Video, 
      "_passthroughFields" -> Map(), 
      "mimeType" -> None, 
      "version" -> 7, 
      "id" -> "https://www.gu.com/something/else", 
      "platform" -> Url)), 
  "_passthroughFields" -> Map(),
  "category" -> Documentary, 
  "metadata" -> Some(Map(
    "tags" -> Some(List()), 
    "_passthroughFields" -> Map(), 
    "commentsEnabled" -> Some(false), 
    "channelId" -> Some("some-awesome-id"))), 
  "title" -> "Amazing title")
```
The implementation of the update method for this structure of maps is just of matter of deciding what the behaviour we want (for example how do we want to handle the update of list). The one included in this PR is not complete, but it works for nested basic types.

The conversion from maps to case classes is very similar, it uses the `FromMap` type class and it mirrors the behaviour of the first one.

The case class to map conversion:
- This takes a case class and returns a `Map[String, Any]`. 
- It's worth noting that scrooge doesn't generate proper case classes from the thrift models. Luckily in the generated models there is an `Immutable` class that can substitute the case classes. This means we need to write one implicit for every thrift class to define the `LabelledGeneric`s for it (see `def anyLabelledGeneric` in `ClassToMap.scala`). I've only added implicits for the media atom, I suggest we auto-generate them in the future.

The map to class conversion:
- This takes a `Map[String, Any]` and returns an `Option[A]` where `A` is the case class we want to convert to.
- A very common reason for which this can fail is that after the update operation a, let's say, field called `date` of type `Long` has been replaced with a `String`. The update itself works fine (considering we do not have any type validation) because of the `Any`. But then when trying to do the conversion, the type class tries to find a label with the name `date` of type `Long`, but fails to find one and returns `None`.

## What we need to do next?
- [ ] write some tests
- [ ] add type validation, the update is not safe at all
- [ ] rewrite the update function to correctly handle all out cases
- [ ] auto-generate the implicits for the `Immutable` class

## Review
This is just a basic implementation and is not ready to be used.
Let me know if you have any questions. Suggestions and contributions are very welcomed.


app/util/ClassToMap.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  // Case for heads that don't have ToMapRec instances (hconsToMapRec1).
+  // Note that we need to use a LowPriority trait to make sure that this instance is
+  // prioritized properly with respect to hconsToMapRec0—if we didn't, the two would have
+  // the same priority and we'd get errors about ambiguous instances.
+  trait LowPriorityToMapRec {
+    implicit def hconsToMapRecNode[K <: Symbol, V, T <: HList]
+    (implicit
+     wit: Witness.Aux[K],
+     tmrT: Lazy[ToMapRec[T]]
```
susiecoleman: It would maybe be nice to give `tmrT` and `tmrH` different names as it makes it a bit hard to read as I have to keep remembering what they are in quite a complex bit of code

app/util/ClassToMap.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  implicit class ToMapRecOps[A](val a: A) extends AnyVal {
+    def toMap[L <: HList]
+    (implicit
+     gen: LabelledGeneric.Aux[A, L],
+     tmr: Lazy[ToMapRec[L]]
+    ): Map[String, Any] = tmr.value(gen.to(a))
+  }
+
+  implicit def mediaAtomLabelledGeneric[L <: HList](implicit gen: LabelledGeneric.Aux[MediaAtom.Immutable, L]): LabelledGeneric.Aux[MediaAtom, L] =
```
susiecoleman: Are we going to need to add to this list of implicit for every new atom type?

app/util/ClassToMap.scala
```diff
@@ -0,0 +1,101 @@
...
+import shapeless.labelled._
+import shapeless._
+
+object ClassToMap {
+  // FROM CASE CLASS TO MAP (http://stackoverflow.com/a/31638390)
+  trait ToMapRec[L <: HList] {
+    def apply(l: L): Map[String, Any]
+  }
+
+  // Case for heads that don't have ToMapRec instances (hconsToMapRec1).
```
susiecoleman: The heads that do have ToMapRec instances get these from the shapeless library right?

app/util/HelperFunctions.scala
```diff
@@ -14,4 +16,34 @@ object HelperFunctions {
...
+  try {
+    Right(updateNestedMap(map, path.split('.').toList, value.as[String]))
+  } catch {
+    case e: AtomAPIError => Left(e)
+  }
+
+  private def updateNestedMap[T](map: Map[String, Any], path: List[String], value: T): Map[String, Any] = path match {
+    case key :: Nil =>
+      // These checks shouldn't be necessary when using proper validation
+      if(map(key).isInstanceOf[Option[Any]]) {
```
susiecoleman: Is this to get round the problem you were having with Options?

app/util/MapToClass.scala
```diff
@@ -0,0 +1,102 @@
...
+
+  trait LowPriorityFromMap {
+    implicit def hconsFromMap1[K <: Symbol, V, T <: HList]
+    (implicit
+     witness: Witness.Aux[K],
+     typeable: Typeable[V],
+     fromMapT: Lazy[FromMap[T]]
+    ): FromMap[FieldType[K, V] :: T] = new FromMap[FieldType[K, V] :: T] {
+      def apply(m: Map[String, Any]): Option[FieldType[K, V] :: T] = for {
+        v <- m.get(witness.value.name)
```
susiecoleman: I know these names are used really locally but I think it might be easier to follow the code if it wasn't single characters as this scala is doing complex stuff...

app/db/AtomWorkshopDB.scala
```diff
@@ -55,5 +58,19 @@ object AtomWorkshopDB {
     transformAtomLibResult(datastore.getAtom(AtomWorkshopDB.buildKey(atomType, id)))
   }
 
+  def updateAtom(datastore: DynamoDataStore[_ >: ExplainerAtom with CTAAtom with MediaAtom], atomType: AtomType, id: String, field: String, value: JsLookupResult) = {
+    for {
+      atom <- transformAtomLibResult(datastore.getAtom(AtomWorkshopDB.buildKey(atomType, id)))
+      atomDataMap = atom.data.asInstanceOf[AtomData.Media].media.toMap
+      updatedAtomData <- update(atomDataMap, field, value)
+      newAtomData <- to[MediaAtom].from(updatedAtomData) match {
```
marialivia16: Good point. This is an example. There are 2 cases here:
1. We only want to update fields inside `data` => we need to find a way do determine what type of data is (e.g. `MediaAtom`). Maybe a pattern match?
2. We want to use the whole atom in which case we can get rid of `asInstanceOf[AtomData.Media]` and use `to[Atom].from(...)`. This is nicer in my opinion because it doesn't add extra logic. But I'm still working separately on how to handle Coproducts in Shapeless (needed because the atom data is a union).

See [here](https://github.com/marialivia16/scrooge-shapeless-maps) a little basic project I'm working on separately to do exactly what we need. There is still some work that I haven't commited yet.

And you can follow [here](http://stackoverflow.com/questions/41985123/unable-to-convert-scrooge-generated-classes-to-maps-using-shapeless) the Stack Overflow question I've asked and partially resolved.

app/util/ClassToMap.scala
marialivia16: Would `toMapRecTail` and `toMapRecHead` be better? (I always read `tomorrow` in my head instead of `toMapRec`...)

app/util/ClassToMap.scala
marialivia16: Yes, most of the times. Or if there is a special cladd we need to define our own implicit. The case when the head has a ToMapRec is when you can recurse on the head. Shapeless know how to generate representations for case classes and basic types.

app/util/ClassToMap.scala
marialivia16: Yes, this is because `scrooge` doesn't generate case classes and we need to tell our type class to use `ClassName.Immutable` instead of `ClassName` when using a `LabelledGeneric`.

On the bright side, all these implicits could be generated using macros.

app/util/HelperFunctions.scala
marialivia16: This is just a temporary check that looks at the value we want to update and if it is an option it will convert the new value to an option as well. Not doing this will result in replacing, for example, an `Option[Long]` with a `Long`. 

Ideally this check shouldn't be necessary anymore when we'll have some type checking.

This `updateNestedMap` method needs a lot of refactoring I'd say. But the changes depend a lot on how we want it to behave and there are things we haven't discussed yet. The way it is now represents just an example.

app/util/MapToClass.scala
marialivia16: Good call.

mchv:
@marialivia16 💯  I am really impressed by your work on this and this is indeed a difficult problem. 

Instead of commenting on the implementation which looks really smart I would like to comment on the initial problem and how by changing a bit the viewpoint you may solve this differently.

If I understand correctly the goal is to provide a generic update endpoint that enables one to update a field of any atom without knowing its type a priori. On the approach you present, because the granularity is on the field you have to deal with a `custom generic deserialisation` for which `shapeless` is probably the best library to based your solution on.

Another approach to solve this problem is to **not** deal about the exact change at this level by requiring a full `resource` to be send on your update endpoint, exacty like a `REST` API.

**What are the benefits?** 
- Serialisation & Deserialisation agnostic.

Your frontend and your backend can rely on the `json` or `thrift` serialisation provided by [content-api-models](https://github.com/guardian/content-api-models/tree/master/json/src/main/scala/com/gu/contentapi/json) but whatever you choose it will not impact your endpoint which don't need to care about it. 

- Conflicts management & resolution simplification

Because you send the entire `resource` it becomes easy to detect incompatible changes on the backend or simply reject older version of the resource if the lastest has been updated since. Note as well that you can generate quite easily a diff between 2 versions for logging purpose.

Hope you will find this interesting.

app/util/ClassToMap.scala
susiecoleman: wooo macros

app/util/ClassToMap.scala
susiecoleman: Yes perfect

app/util/ClassToMap.scala
marialivia16: [This](https://github.com/travisbrown/scrooge-circe-demo/blob/master/src/main/scala/demo/package.scala#L18) looks like a good example.

tomrf1:
Wow, this looks very interesting!

Out of interest, why can't you do your update against a JSON-representation of your atom object? If the type or path provided by the update request is wrong then you'd find out when you try to decode back to the class.

app/db/AtomWorkshopDB.scala
```diff
@@ -55,5 +58,19 @@ object AtomWorkshopDB {
     transformAtomLibResult(datastore.getAtom(AtomWorkshopDB.buildKey(atomType, id)))
   }
 
+  def updateAtom(datastore: DynamoDataStore[_ >: ExplainerAtom with CTAAtom with MediaAtom], atomType: AtomType, id: String, field: String, value: JsLookupResult) = {
```
philmcmahon: Good point. We need to document this I suspect, unless we can find a way to represent the type more nicely.

marialivia16:
@tomrf1 Thanks!

We talked a lot about how to approach this problem and we also considered JSON. We found that both solutions faces the same issues. For example handling optional values. By implementing this we get a little more flexibility in deciding how to handle some cases.

I guess we'll keep this solution in a test phase and see how it behaves and if it's something we want. I'm separately working on [this](https://github.com/marialivia16/scrooge-shapeless-maps) repo - it's not as advanced as this PR, but I plan to make it even more generic in the future. In case we decide not to use it in this project it will at least live somewhere else.

app/util/HelperFunctions.scala
```diff
@@ -14,4 +16,34 @@ object HelperFunctions {
...
+        map.updated(key, Some(value))
+      } else if(value.getClass != map(key).getClass) {
+        Logger.error("Types not matching when trying to update.")
+        throw new WrongTypeFound(found = value.getClass.getCanonicalName, expected = map(key).getClass.getCanonicalName)
```
philmcmahon: So far I've tried to avoid throwing errors in the API, instead using eithers and returning them as a json object to the client. I can tidy this up whilst doing the merge!

tomrf1:
@marialivia16 following on from our conversation earlier, it's worth exploring the use of circe for a simpler solution, in particular its [optics](https://circe.github.io/circe/optics.html) library. I don't expect there to be any issues with optional types.

marialivia16:
@tomrf1 Here's a PR with the circe work https://github.com/guardian/atom-workshop/pull/40

marialivia16:
I'm closing this PR as we have a better implementation instead: https://github.com/guardian/atom-workshop/pull/40

PR #36, opened 2017-02-08
Hot Reloading dev workflow

clloyd:
This adds Hot reloading!

Start the app with `./scripts/client-dev.sh` and 🔥  reload away.

This required some changes to the structure of the frontend

1. There is now a BaseApp.js component which replaces the router, this allows all react component changes to bubble up to a single location, and allows hot reloading to happen at that level
2. I re-created the `rootReducer` file for the same reason, this allows us to monitor a single module and receive all reducer changes though them bubbling to that module. 

See comments in code for interesting implementation details


.babelrc
```diff
@@ -1,5 +1,17 @@
 // .babelrc
 {
-  "presets": ["es2015", "react"],
-  "plugins": ["transform-object-assign", "transform-class-properties"]
+  "presets": [
+    ["es2015", {"modules": false}],
```
clloyd: `{"modules": false}` removes the es6 module -> commonjs transform.

.babelrc
```diff
@@ -1,5 +1,17 @@
...
+    "react"
+  ],
+  "plugins": [
+    "transform-object-assign",
+    "transform-class-properties",
+    "react-hot-loader/babel"
+  ],
+  "env": {
+    "test": {
+      "plugins": ["transform-es2015-modules-commonjs"]
```
clloyd: Unfortunately jest needs this transformation, so it's re-added in test

public/js/app.js
```diff
@@ -32,8 +34,20 @@ store.dispatch({
...
+      {Component}
+    </AppContainer>,
+    document.getElementById('react-mount')
+  );
+};
+
+
+render(<BaseApp store={store} />);
+
+if (module.hot) {
```
clloyd: This is now required... we're listening for the hot module changes and triggering a re-render

public/js/util/configureStore.js
```diff
@@ -19,5 +11,14 @@ const createStoreWithMiddleware = compose(
 )(createStore);
 
 export default function configureStore(initialState) {
-  return createStoreWithMiddleware(rootReducer, initialState);
+
+  const store = createStoreWithMiddleware(rootReducer, initialState);
+
+  if (module.hot) {
```
clloyd: This allows us to trigger differently depending on what's changed... here we trigger a reduce replacement (rather than a re-render)

conf/webpack-devserver.conf.js
```diff
@@ -0,0 +1,65 @@
...
+      path: resolve(__dirname, '..', 'public', 'build'),
+      publicPath: 'https://atomworkshop-assets.local.dev-gutools.co.uk/assets/build/'
+    },
+
+    devServer: {
+      hot: true,
+      contentBase: resolve(__dirname, '..', 'public', 'build'),
+      publicPath: 'https://atomworkshop-assets.local.dev-gutools.co.uk/assets/build/',
+      port: 9051,
+      public: "atomworkshop-assets.local.dev-gutools.co.uk"
```
akash1810: minor: can we have this string as a `const` for DRYness?

package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "build-dev": "webpack ./public/js/app.js ./public/build/app.js --config ./conf/webpack.conf.js --watch",
     "build": "webpack -p ./public/js/app.js ./public/build/app.js --config ./conf/webpack.conf.js",
+    "hmr": "webpack-dev-server --config ./conf/webpack-devserver.conf.js",
```
akash1810: what does `hmr` stand for? hot module reload?

clloyd: yeah

conf/webpack-devserver.conf.js
clloyd: I've no objection to doing it for DRY reasons... but I actually think it's clearer it all being written out. Compared to the same file in media atom maker https://github.com/guardian/media-atom-maker/blob/master/build_config/webpack.devserver.conf.js

clloyd: Especially as webpack-dev-server confs are absolutely insane... `publicPath` vs `public` etc etc

PR #38, opened 2017-02-13
Add Linter

clloyd:
Lets enforce some javascript standards! woo!

This adds `eslint` and `eslint-plugin-react` with the recommended rules, most are pretty regular... but important ones that required the most changes:
 - PropTypes are REQUIRED for all props, this is a really good change... 
 - `console.anything` is disallowed, this is also good... we should have this in an util so sentry is easier to add.

once merged will enable in the build...


public/js/util/logger.js
```diff
@@ -0,0 +1,11 @@
+/* eslint no-console: "off" */
+
+//TODO implement sentry
+
+export function logInfo(...message) {
+  console.log.apply(this, message);
```
akash1810: why `apply`?

clloyd: So you can still do `console.log(a, b, [c,d])`

clloyd:
@akash1810 we're using teamcity for this build... build steps aren't committed unfortunately.

PR #39, opened 2017-02-13
Added createAtom functionality

clloyd:
This adds a working createAtom functionality. Only functional with CTA and Explainers while work is being done for non-complete atoms. I've modified the create atom endpoint so it returns the full atom on a successful create

Important things from this... 

 - `fetch` requires us to pass `credentials: 'same-origin'`
 - we need to `res.json()` the result

### DIFF
--- a/public/js/components/AtomCreate/AtomCreateGenericInfo.js
+++ b/public/js/components/AtomCreate/AtomCreateGenericInfo.js
@@ -4,13 +4,14 @@ import {getAtomByType} from '../../constants/atomData';
 import {AtomTypeCard} from '../AtomTypeCard/AtomTypeCard.js';
 import FormFieldTextInput from '../FormFields/FormFieldTextInput';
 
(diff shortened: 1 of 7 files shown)


public/js/components/AtomCreate/AtomCreateGenericInfo.js
```diff
@@ -53,9 +54,28 @@ export class AtomCreateGenericInfo extends React.Component {
...
 }
+
+//REDUX CONNECTIONS
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+import * as createAtomActions from '../../actions/AtomActions/createAtom.js';
+
+function mapStateToProps(state) {
+  return {
+    config: state.config
```
ShaunYearStrong: Is this required atm? If there's no additional overhead from having this then might as well leave it but just wondering.

PR #40, opened 2017-02-13
Generic update endpoint using Circe

marialivia16:
The first attempt to solve this using Shapeless can be found here: https://github.com/guardian/atom-workshop/pull/35

At @tomrf1 's suggestion, here's the solution using Circe. But not without any issues.

Note, I am not using [optics](https://circe.github.io/circe/optics.html) because it assumes I can do something like `atom.metadata.channelId`. My path is a string, it doesn't do anything clever.

Also, the work in this PR only knows how to update fields of type `String` or `Option[String]` at the moment. This is because the body of the request looks like this: `{"value": "some value to update"}`. When I retrieve this JSON I have to convert it to something (`parsedJson.hcursor.downField("value").as[String]`), but I don't know a priori the type. And this is not the only place where I need to know the type of the new value. When I am doing the actual update I have to use the appropriate method: `cursor.withFocus(_.mapString(_ => newValue))`.

For now, I'm assuming all fields are strings and I'm investigating a way around it.

Have a look 👀 and leave your suggestions 💬 

**EDIT**
The endpoint is now `PATCH /api/preview/:atomType/:id`
The request body looks like this:
```json
{
  "data": {
    "media": {
       "activeVersion": 1111
    }
  }
}
```


tomrf1:
Excellent!
First, does the request really have to have the form `{"value": "some value to update"}`?
If the request payload could instead define the path as well then it could really help!
E.g. If the client sends `atom: { metadata: { channelId: "newId" } }`
Then you could use `currentJson.deepMerge(updateJson)`, and it'll do everything for you.

If that's not possible then it might still be preferable to construct a separate json object and then use `deepMerge`. Perhaps something like:

```
def construct(path: List[String], value: Json): List[(String, Json)] = {
  path match {
    case x :: Nil => List((x, value))
    case x :: xs => List((x, Json.fromFields(construct(xs, value))))
    case Nil => Nil
  }
}
    
val originalJson: Json = //encode scrooge class

val path = "atom.metadata.channelId".split('.').toList

val value: Json = parse("""{"value": "some value to update"}""").toOption.flatMap(_.hcursor.downField("value").success.map(_.value)).get

val updateJson = Json.fromFields(construct(path, value))

val newJson = originalJson.deepMerge(updateJson)
```

marialivia16:
Thanks, @tomrf1 ! I'm looking at this today.

app/db/AtomWorkshopDB.scala
```diff
@@ -37,24 +35,51 @@ object AtomWorkshopDB {
...
-      defaultHtml = buildDefaultHtml(atomType, currentVersion.data, Some(currentVersion.defaultHtml)),
-      data = newAtom.data
-    )
+  private def updateAtomInDatastore(datastore: DynamoDataStore[_ >: ExplainerAtom with CTAAtom with MediaAtom], atom: Atom): Either[AtomAPIError, Unit] = {
     try {
-      val result = datastore.updateAtom(updatedAtom)
-      Logger.info(s"Successfully updated atom of type ${atomType.name} with id ${currentVersion.id}")
+      val result = datastore.updateAtom(atom)
+      Logger.info(s"Successfully updated atom of type ${atom.atomType.name} with id ${atom.id}")
       Right(transformAtomLibResult(result))
```
susiecoleman: Is this wrapping an `Either` in an `Either`?

app/db/AtomWorkshopDB.scala
```diff
@@ -37,24 +35,51 @@ object AtomWorkshopDB {
...
+  }
+
+  private def createAtomFromUpdatedAtom(atom: Atom, updatedAtom: Atom, user: User): Atom =
+    atom.copy(
+      contentChangeDetails = buildContentChangeDetails(user, Some(atom.contentChangeDetails), updateLastModified = true),
+      defaultHtml = buildDefaultHtml(atom.atomType, updatedAtom.data, Some(atom.defaultHtml)),
+      data = updatedAtom.data
+    )
+
+  def updateAtom(datastore: DynamoDataStore[_ >: ExplainerAtom with CTAAtom with MediaAtom],
```
susiecoleman: If you wanted to be really picky you could call this `updateEntireAtom` to match the controller

app/util/HelperFunctions.scala
```diff
@@ -39,15 +38,33 @@ object HelperFunctions {
     Either.cond(body.isDefined, body.get, BodyRequiredForUpdateError)
   }
 
-  def parseAtomJson(atomJson: String): Either[AtomAPIError, Atom] = {
-    Logger.info(s"Parsing atom json: ${atomJson}")
+  def parseStringToAtom(atomString: String): Either[AtomAPIError, Atom] = {
```
susiecoleman: It doesn't need to be in this PR (or at all) but at some point we could add some tests around the helper functions taking strings as params as that would be good documentation of what a valid string looks like

app/db/AtomWorkshopDB.scala
```diff
@@ -37,24 +35,51 @@ object AtomWorkshopDB {
...
+  def updateAtomByPath(datastore: DynamoDataStore[_ >: ExplainerAtom with CTAAtom with MediaAtom],
+                       atomType: AtomType,
+                       user: User,
+                       currentJson: Json,
+                       newJson: Json): Either[AtomAPIError, Unit]  = {
+
+    val updatedAtomJson: Json = currentJson.deepMerge(newJson)
+
+    val updatedAtom = for {
+      atom <- parseStringToAtom(currentJson.toString)
```
tomrf1: You don't need to stringify the `Json`.
Instead you can do: `currentJson.as[Atom]`

philmcmahon:
Finally got round to look at this. Nice work!!! 💯

build.sbt
```diff
@@ -13,11 +13,11 @@ libraryDependencies ++= Seq(
   "com.gu"                   %% "atom-manager-play"           % atomLibVersion,
   "com.gu"                   %% "atom-publisher-lib"          % atomLibVersion,
   "com.gu"                   %% "configuration-magic-core"    % "1.3.0",
+  "com.gu"                   %  "fezziwig"                    % "0.1.1",
```
mchv: should be  ` "com.gu"                   %%  "fezziwig"                    % "0.2",`

PR #41, opened 2017-02-14
Pass require prop into Modal test

clloyd:
Dismiss is now a required prop for the `<Modal />` component, this passes through the prop in the test (even though it's not used)

### DIFF
--- a/public/js/components/Utilities/Modal.spec.js
+++ b/public/js/components/Utilities/Modal.spec.js
@@ -5,7 +5,7 @@ import {shallow} from 'enzyme';
 
 test('Should render null when closed', () => {
   const component = renderer.create(
-    <Modal isOpen={false}><div>Modal Contents</div></Modal>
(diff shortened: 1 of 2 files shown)


public/js/components/Utilities/Modal.spec.js
```diff
@@ -5,15 +5,15 @@ import {shallow} from 'enzyme';
 
 test('Should render null when closed', () => {
   const component = renderer.create(
-    <Modal isOpen={false}><div>Modal Contents</div></Modal>
+    <Modal isOpen={false} dismiss={() => {}}><div>Modal Contents</div></Modal>
```
ShaunYearStrong: I like that this is explicitly an empty function, but should we make a new jest function and pass that in?

clloyd: So further down that's what we do, when we're actually testing if those functions are being called... I quite like that it looks empty when we're not testing anything to do with it.

PR #42, opened 2017-02-14
Require panda-session be loaded through babel-loader

clloyd:
Fixing the build errors, #39 actually used panda-session (it was previously being excluded), when it hit UglifyJS the ES6-ness caused the build to fail. This ensures it is passed through babel

### DIFF
--- a/conf/webpack-devserver.conf.js
+++ b/conf/webpack-devserver.conf.js
@@ -1,3 +1,6 @@
+/* global module:false, __dirname:false */
+
+
 const { resolve } = require('path');
 const webpack = require('webpack');
 
@@ -28,6 +31,11 @@ module.exports = {
             {
                 test:    /\.js$/,
                 exclude: /node_modules/,
+                loaders: ['babel-loader']
+            },
+            {
+                test:    /\.js$/,
+                include: [resolve(__dirname, "../node_modules/panda-session")],
                 loaders: ['babel-loader']
             },
             {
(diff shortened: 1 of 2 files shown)


conf/webpack-devserver.conf.js
```diff
@@ -31,6 +34,11 @@ module.exports = {
                 loaders: ['babel-loader']
             },
             {
+                test:    /\.js$/,
+                include: [resolve(__dirname, "../node_modules/panda-session")],
```
ShaunYearStrong: Can't this be included in the block above?

conf/webpack.conf.js
```diff
@@ -12,6 +14,11 @@ module.exports = {
                 loaders: ['babel-loader']
             },
             {
+                test:    /\.js$/,
+                include: [path.resolve(__dirname, "../node_modules/panda-session")],
```
ShaunYearStrong: As above.

conf/webpack-devserver.conf.js
clloyd: Unfortunately not... without clever regex-foo, simply adding include will cause everything else to be excluded, and I don't really want to have to put the public path in the include... also like the separation of a separate rule, can remove once uglify get it's act together

ShaunYearStrong: Fair enough. I thought webpack would do the clever regex-foo

PR #43, opened 2017-02-14
Get atom automatically on edit page, add Update action

ShaunYearStrong:


public/js/components/AtomEdit/AtomEdit.js
```diff
@@ -1,9 +1,54 @@
...
+  static propTypes = {
+    routeParams: PropTypes.shape({
+      atomType: PropTypes.string.isRequired,
+      id: PropTypes.string.isRequired
+    }).isRequired,
+    atomActions: PropTypes.shape({
+      getAtom: PropTypes.func.isRequired,
+      updateAtom: PropTypes.func.isRequired
+    }).isRequired,
+    atom: PropTypes.object
```
clloyd: Can we define a PropType for this? only needs to be the base items e.g. id and type

ShaunYearStrong: Sure

ShaunYearStrong:
@clloyd Added that here: https://github.com/guardian/atom-workshop/blob/c2acb64308c2c9a5905859a4eb2c9ef860b2856b/public/js/components/AtomEdit/AtomEdit.js#L3-L9 I didn't add `contentChangeDetails` for now as I wasn't sure what the required fields were for an atom.

Also, as this is a pretty standard model, perhaps we should move the `atomPropType` const out to a constant file?

PR #44, opened 2017-02-15
Possible direction for Managed Form

clloyd:
cc @ShaunYearStrong 

This is the basis for an easy way to generate editor forms.


public/js/components/AtomEdit/AtomEdit.js
```diff
@@ -31,12 +32,21 @@ class AtomEdit extends React.Component {
...
-        <p>Editing atom: </p>
-        <button className="btn" type="button" onClick={this.updateAtom}>Update atom</button>
-      </dev>
-    );
+
+    if (!this.props.atom) {
+      return <div>Loading...</div>;
+    }
+
+    const atomType = this.props.atom.atomType.toLowerCase();
```
ShaunYearStrong: I know this is linked to the data coming back but I'm not a fan. Maybe we could add a note?

public/js/components/ManagedEditor/index.js
```diff
@@ -0,0 +1,2 @@
+export {ManagedField} from './ManagedEditorField';
```
ShaunYearStrong: I like this 👍

public/js/components/AtomEdit/AtomEdit.js
clloyd: I completely agree, will add a note

PR #45, opened 2017-02-15
[WIP] - managed field validation

ShaunYearStrong:


public/js/components/AtomEdit/CustomEditors/CTAEditor.js
```diff
@@ -18,19 +18,19 @@ export class CTAEditor extends React.Component {
     return (
       <div className="editor editor-cta">
         <ManagedForm data={this.props.atom} updateData={this.props.onUpdate}>
-          <ManagedField fieldLocation="data.url" name="Link Url" isRequired={true}>
+          <ManagedField fieldLocation="data.url" name="Link Url" fieldId="url" isRequired={true}>
```
clloyd: Does this fieldId need to be provided? Can we get a unique if from the react component (might not be human readable)

public/js/components/ManagedEditor/ManagedEditorForm.js
```diff
@@ -12,5 +12,5 @@ export const ManagedForm = (props) => {
 
 ManagedForm.propTypes = {
   updateData: PropTypes.func.isRequired,
-  children: PropTypes.element.isRequired
+  children: PropTypes.array.isRequired
```
clloyd: I fixed this in my branch by setting to
```
  children: PropTypes.oneOfType([
    PropTypes.element,
    PropTypes.arrayOf(PropTypes.element)
  ])
```

public/js/util/validators.js
```diff
@@ -1,9 +1,23 @@
...
+
+
 export const isHttpsUrl = (value) => {
   const stringValue = typeof value === 'string' ? value : '';
-
   if (stringValue.match(/^(https?:\/\/)?([\da-z\.-]+)\.([a-z\.]{2,6})([\/\w \.-]*)*\/?$/)) {
     return true;
   } else {
-    return "Not a HTTPS url";
+    return {
```
clloyd: :+1:

public/js/components/AtomEdit/CustomEditors/CTAEditor.js
ShaunYearStrong: that'd be fine. I'm not actually using it atm but thought a UID would be useful if we had a list of all field errors.

public/js/util/validateField.js
```diff
@@ -0,0 +1,25 @@
...
+    });
+  }
+
+  // Custom validators
+  if(customValidation) {
+    customValidation.forEach((validator) => {
+      const result = validator(fieldValue);
+
+      if (result !== true) {
+        errors.push(result);
```
clloyd: Do we want to throw an error here if it's not true... but also doesn't provide an object we recognise?

ShaunYearStrong: Good idea, it'd be good to catch it here rather that it being unrecognised on the field.

public/js/components/AtomEdit/CustomEditors/CTAEditor.js
ShaunYearStrong: @clloyd Hmm, looks like for now, the key is being set to `.0` for all of them, so it's not unique. This was more a change to fix the spaces in the `id`

PR #46, opened 2017-02-15
Add delete functionality

susiecoleman:
It is now possible to delete atoms from live and preview. The following logic applies:
1. Deleting from live - Atom is removed from the live table and left in preview
2. Deleting from preview - Check if atom is live. If it is do not delete from live or preview. If atom is only in preview delete it.


app/db/AtomWorkshopDB.scala
```diff
@@ -38,7 +48,11 @@ object AtomWorkshopDB {
   }
 
   def getAtom(datastore: DynamoDataStore[_ >: ExplainerAtom with CTAAtom with MediaAtom], atomType: AtomType, id: String) = {
-    transformAtomLibResult(datastore.getAtom(AtomWorkshopDB.buildKey(atomType, id)))
+    transformAtomLibResult(datastore.getAtom(buildKey(atomType, id)))
+  }
```
marialivia16: 😃 No need for `{` 
Same for `deleteAtom`

susiecoleman: Good catch 🐟

PR #51, opened 2017-02-16
Add recipe steps (with array wrapper)

clloyd:
@ShaunYearStrong 

screen shot 2017-02-16 at 12 08 55

Ugly... but the building blocks of an array handler around fields. Also removed a lot of the propType warnings


public/js/components/AtomEdit/CustomEditors/RecipeFields/Servings.js
```diff
@@ -8,16 +8,16 @@ import {logError} from '../../../../util/logger';
 export class RecipeServings extends React.Component {
 
   static propTypes = {
-    fieldLabel: PropTypes.string.isRequired,
-    fieldName: PropTypes.string.isRequired,
+    fieldLabel: PropTypes.string,
```
ShaunYearStrong: :(

PR #55, opened 2017-02-16
Add ami update dependency to deploy.

susiecoleman:
Adding an ami update dependency so that the ami is updated before deployment.


jennysivapalan:
👍 🆙

PR #56, opened 2017-02-16
Add explainer atom form

ShaunYearStrong:
First pass.

Improvements:

- [x] Add Rich text editor(Scribe?) to body

picture 91


public/js/components/AtomEdit/CustomEditors/ExplainerEditor.js
```diff
@@ -0,0 +1,33 @@
+import React, {PropTypes} from 'react';
+import {ManagedForm, ManagedField} from '../../ManagedEditor';
+import FormFieldTextInput from '../../FormFields/FormFieldTextInput';
+import FormFieldsScribeEditor from '../../FormFields/FormFieldScribeEditor';
+import {ExplainerDisplayTypes} from './ExplainerFields/DisplayTypes';
+
+export class ExplainerEditor extends React.Component {
+
+  static propTypes = {
```
susiecoleman: Just wanted to check I've understood PropTypes. Is this saying that the ExplainerEditor must be passed: 
1. An atom which contains the key `type` which must have a string value
2. A function called `onUpdate`

ShaunYearStrong: That's correct! You can omit the `isRequired` too, if the prop is optional.

PR #59, opened 2017-02-16
Split checkboxes into groups and singles

ShaunYearStrong:
cc @clloyd

### DIFF
--- a/public/js/components/FormFields/FormFieldCheckbox.js
+++ b/public/js/components/FormFields/FormFieldCheckbox.js
@@ -12,15 +12,15 @@ export default class FormFieldCheckbox extends React.Component {
     onUpdateField: PropTypes.func.isRequired
   }
 
-  onUpdate = () => {
-    this.props.onUpdateField(this.props.fieldName);
+  onUpdate = (e) => {
+    this.props.onUpdateField(e.target.value);
   }
 
   render() {
     return (
       <div className="form__group form__group--checkbox">
         {this.props.fieldLabel ? <label className="form__label" htmlFor={this.props.fieldName}>{this.props.fieldLabel}</label> : false}
-        <input className="form__checkbox" type="checkbox" checked={this.props.fieldValue} name={this.props.fieldName} onChange={this.onUpdate} />
+        <input className="form__checkbox" type="checkbox" checked={this.props.fieldValue} name={this.props.fieldName} value={this.props.fieldValue} onChange={this.onUpdate} />
         {!this.props.fieldLabel ? <span className="form__label form__label--checkbox">{this.props.fieldName}</span> : false}
         <ShowErrors errors={this.props.fieldErrors}/>
       </div>
--- a/public/js/components/FormFields/FormFieldCheckbox.js
+++ b/public/js/components/FormFields/FormFieldCheckbox.js
@@ -5,38 +5,25 @@ import { errorPropType } from '../../constants/errorPropType';
 export default class FormFieldCheckbox extends React.Component {
 
   static propTypes = {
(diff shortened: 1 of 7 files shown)


public/js/components/FormFields/FormFieldCheckbox.js
```diff
@@ -5,38 +5,25 @@ import { errorPropType } from '../../constants/errorPropType';
...
-        <input className="form__checkbox" type="checkbox" checked={value.selected} name={this.props.fieldName} value={value} onChange={this.onUpdate} />
-        <span className="form__label form__label--checkbox">{value.id}</span>
-      </div>
-    );
   }
 
-  renderCheckboxes() {
-    return this.props.fieldValue.map(this.renderCheckbox, this);
+  onUpdate = () => {
+    this.props.onUpdateField(this.props.fieldName);
```
clloyd: Should this not return the bool?

clloyd: Thinking that if we had something like `isExpandable` as a checkbox... I'd expect it to return `true` `false` ?

ShaunYearStrong: @clloyd True, but I was thinking of the data needed for the group, would that not be a list of the names of checkboxes which are true?

public/js/components/FormFields/FormFieldCheckboxGroup.js
```diff
@@ -0,0 +1,68 @@
...
+
+  isChecked = (checkValue) => {
+    return this.props.fieldValue.includes(checkValue);
+  }
+
+  renderCheckbox(fieldName, i) {
+    const updateFn = (newValue) => {
+      let newFieldValue = [];
+      if(newValue) {
+        newFieldValue = this.props.fieldValue.concat(fieldName);
```
clloyd: Doesn't concat require an array to be passed to it?

clloyd: Also should we check is it's already in there?

ShaunYearStrong: Yep, and that should happen as the true/false will _totally_ work. But I'll add a check just in case haha

public/js/components/FormFields/FormFieldCheckboxGroup.js
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  isChecked = (checkValue) => {
+    return this.props.fieldValue.includes(checkValue);
+  }
+
+  renderCheckbox(fieldName, i) {
+    const updateFn = (newValue) => {
+      let newFieldValue = [];
+      if(newValue && !this.props.fieldValue.includes(newValue)) {
```
philmcmahon: Could you use !isChecked here instead?

philmcmahon:
Other than possible simplifications above, 👍

public/js/components/FormFields/FormFieldCheckboxGroup.js
```diff
@@ -0,0 +1,68 @@
...
+    fieldValue: PropTypes.array.isRequired,
+    checkValues: PropTypes.array.isRequired,
+    fieldErrors: PropTypes.arrayOf(errorPropType),
+    onUpdateField: PropTypes.func.isRequired
+  };
+
+  addOrRemoveValue = (fieldName) => {
+    let newFieldValue = [];
+
+    if(this.props.fieldValue.includes(fieldName)) {
```
ShaunYearStrong: It's unused, I left it here accidentally. Good spot!

PR #62, opened 2017-02-21
Add publishing to kinesis and publish route

philmcmahon:
This change adds an object with live/preview kinesis publishers, which are then used in the various controllers to push changes to capi. The change also adds a publish endpoint, which checks if something exists in draft, and then pushes it to the live dynamo db table.

We also tidied up the super long AtomWorkshopDatastore type.

This PR uses this branch: https://github.com/guardian/atom-maker/pull/11 which is unfinished, but does the job we need - to get the update operation of the atom maker to return the updated atom. When @susiecoleman has finished we can update the dependency to point at the release version of the atom maker rather than a snapshot of that branch.

### DIFF
--- a/app/config/Config.scala
+++ b/app/config/Config.scala
@@ -29,14 +29,14 @@ object Config extends AwsInstanceTags {
   }
   val config = ConfigurationMagic(appName, configMagicMode).load
 
-  def getOptionalProperty[T](path: String, getVal: String => T) = {
+  def getOptionalProperty[T](path: String, getVal: String => T): Option[T] = {
     if (config.hasPath(path)) Some(getVal(path))
     else None
   }
 
-  def getPropertyIfEnabled(enabled: Boolean, path: String) =
+  def getPropertyIfEnabled(enabled: Boolean, path: String): String =
     if (enabled) getOptionalProperty(path, config.getString).getOrElse(sys.error(s"Property $path is required"))
-    else ""
+    else s"feature requiring $path is disabled"
 
   val kinesisEnabled = getOptionalProperty("aws.kinesis.publish.enabled", config.getBoolean).getOrElse(true)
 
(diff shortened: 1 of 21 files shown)


app/config/Config.scala
```diff
@@ -28,12 +29,18 @@ object Config extends AwsInstanceTags {
...
-  def getOptionalProperty[T](path: String, getVal: String => T) = {
+  def getOptionalProperty[T](path: String, getVal: String => T): Option[T] = {
     if (config.hasPath(path)) Some(getVal(path))
     else None
   }
 
-  val elkKinesisStream = config.getString("elk.kinesis.stream")
+  def getPropertyIfEnabled(enabled: Boolean, path: String): String =
+    if (enabled) getOptionalProperty(path, config.getString).getOrElse(sys.error(s"Property $path is required"))
+    else s"feature requiring $path is disabled"
```
clloyd: Should this be an error of some sort?

app/controllers/App.scala
```diff
@@ -48,10 +50,24 @@ class App(val wsClient: WSClient, val atomWorkshopDB: AtomWorkshopDBAPI) extends
         atomType <- validateAtomType(atomType)
         ds <- AtomDataStores.getDataStore(atomType, Preview)
         atom <- atomWorkshopDB.createAtom(ds, atomType, req.user)
+        _ <- sendKinesisEvent(atom, previewAtomPublisher, EventType.Update)
       } yield atom
     }
   }
 
+  def publishAtom(atomType: String, id: String) = AuthAction { req =>
```
clloyd: I think this is a bit more complicated that this unfortunately (chatted with Phil in person)

app/config/Config.scala
philmcmahon: I don't think so - it will only be used if the feature has been disabled via a config switch which isn't an error exactly

app/controllers/App.scala
philmcmahon: Updated!

clloyd:
👍

PR #69, opened 2017-02-27
Pass title through on create and allow editing

clloyd:
screen shot 2017-02-27 at 15 25 49

Pass through title on create and allow editing on the edit screen

### DIFF
--- a/public/js/components/AtomEdit/AtomEdit.js
+++ b/public/js/components/AtomEdit/AtomEdit.js
@@ -2,14 +2,9 @@ import React, { PropTypes } from 'react';
 import {CTAEditor} from './CustomEditors/CTAEditor';
 import {RecipeEditor} from './CustomEditors/RecipeEditor';
 import {ExplainerEditor} from './CustomEditors/ExplainerEditor';
+import AtomEditHeader from './AtomEditHeader';
 
-const atomPropType = PropTypes.shape({
-  id: PropTypes.string.isRequired,
-  atomType: PropTypes.string.isRequired,
-  labels: PropTypes.array.isRequired,
-  defaultHtml: PropTypes.string.isRequired,
-  data: PropTypes.object.isRequired
-});
+import {atomPropType} from '../../constants/atomPropType.js';
 
 class AtomEdit extends React.Component {
 
@@ -36,11 +31,7 @@ class AtomEdit extends React.Component {
     this.props.atomActions.updateAtom(newAtom);
   }
 
-  render () {
-
-    if (!this.props.atom) {
-      return <div>Loading...</div>;
-    }
+  renderSpecificEditor () {
 
     //TODO: This is brittle, can we improve?
     const atomType = this.props.atom.atomType.toLowerCase();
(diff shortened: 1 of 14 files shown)


public/js/components/AtomEdit/AtomEdit.js
```diff
@@ -36,11 +31,7 @@ class AtomEdit extends React.Component {
...
 
-  render () {
-
-    if (!this.props.atom) {
-      return <div>Loading...</div>;
-    }
+  renderSpecificEditor () {
 
     //TODO: This is brittle, can we improve?
     const atomType = this.props.atom.atomType.toLowerCase();
```
shaundillon: I think we can get this from the key inside `data` now. Not great but at least it's the same case

public/js/components/AtomEdit/AtomEditHeader.js
```diff
@@ -0,0 +1,51 @@
...
+        </div>
+        <div className="atom-card__details">
+          <h3 className="atom-card__heading">{this.props.atom.atomType} Atom</h3>
+          <ManagedForm data={this.props.atom} updateData={this.props.onUpdate}>
+            <ManagedField fieldLocation="title" name="Title:">
+              <FormFieldTextInput/>
+            </ManagedField>
+          </ManagedForm>
+          <div className="atom-card__description">
+            <div><b>Created: </b>{this.renderDate('created')}</div>
```
shaundillon: I think this should be a `<ul>`

public/js/components/AtomEdit/AtomEdit.js
clloyd: not sure I know what you mean?

public/js/components/AtomEdit/CustomEditors/RecipeEditor.js
```diff
@@ -24,35 +24,32 @@ export class RecipeEditor extends React.Component {
...
-            </FormFieldArrayWrapper>
-          </ManagedField>
-          <ManagedField fieldLocation="data.recipe.images" name="Images">
-            <FormFieldArrayWrapper>
-              <FormFieldImageSelect gridUrl={this.props.config.gridUrl}/>
-            </FormFieldArrayWrapper>
-          </ManagedField>
-        </ManagedForm>
-      </div>
+      <ManagedForm data={this.props.atom} updateData={this.props.onUpdate}>
```
shaundillon: Really like that these are just a managed form now.

public/js/services/AtomsApi.js
```diff
@@ -4,44 +4,46 @@ export default {
 
   getAtom: (atomType, atomId) => {
     return pandaFetch(
-      `/api/preview/${atomType}/${atomId}`,
-      {
-        method: 'get',
-        credentials: 'same-origin'
-      }
+      new Request(
```
shaundillon: These didn't seem to work when I changed polyfills before, so I was just sending them separately. `pandaFetch.js` may need it's params changing.

public/js/components/AtomEdit/AtomEdit.js
shaundillon: the object immediately inside `data` has the same key as the type of the atom. Not the nicest but it'll get around the use of `toLowerCase`

PR #71, opened 2017-02-27
Saved indicator date

shaundillon:
picture 112


public/js/components/Header/Header.js
```diff
@@ -26,14 +27,20 @@ class Header extends React.Component {
   }
 
   renderSaveState = () => {
+    const dateNow = Date.now();
+    const lastModified = this.props.atom.contentChangeDetails.lastModified ? this.props.atom.contentChangeDetails.lastModified.date : dateNow;
+
+    const timeSinceLastModified = distanceInWords(dateNow, lastModified, {addSuffix: true});
+
+    console.log(dateNow, lastModified, timeSinceLastModified);
```
clloyd: Console.log also... should it default to date created when there's not a modified date?

PR #72, opened 2017-02-28
[EXPERIMENT] - Debounce redux updates while maintaining state

shaundillon:
based on [this article](https://medium.com/@justintulk/debouncing-reacts-controlled-textareas-w-redux-lodash-4383084ca090#.supzyawqv).

Stores the value in local state, sends this to the updateFunction after 250ms of no function calls.

##Problems

- If Redux dispatch, the state will be intact, but won't get saved any more.
- Use of local state inside managed fields feels flimsy, no direct reflection of the atom in the redux store

## Advantages
- Abstracts debouncing from the action creators, keeping them intact for use elsewhere

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 target
 .idea
+.vscode
 .DS_Store
 logs/
 node_modules/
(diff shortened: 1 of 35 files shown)


.gitignore
```diff
@@ -1,5 +1,6 @@
 target
 .idea
+.vscode
```
shaundillon: Yep, I'm using VScode now...

philmcmahon:
👍

PR #74, opened 2017-02-28
Basic Story QuestionsEditor and Tag Picker

clloyd:
This adds:
 - Preview CAPI proxy and Support controller (config added to all ENVs)
 - Full Support for the StoryQuestions atom server side
 - Support for Story Question creation
 - Tag Picker without styling

cc @philmcmahon @shaundillon


app/controllers/Support.scala
```diff
@@ -0,0 +1,43 @@
...
+    httpClient.setConnectTimeout(5, TimeUnit.SECONDS)
+
+    val resp = for {
+      capiPreviewUser <- Config.capiUsername
+      capiPreviewPassword <- Config.capiPassword
+      capiUrl <- Config.capiPreviewUrl
+    } yield {
+      val url = s"$capiUrl/$path?${request.rawQueryString}"
+
+      println(url)
```
LATaylor-guardian: Is this meant to be here?

app/controllers/Support.scala
```diff
@@ -0,0 +1,43 @@
...
+import com.squareup.okhttp.{Credentials, OkHttpClient, Request}
+import config.Config
+import java.util.concurrent.TimeUnit
+
+
+
+class Support(val wsClient: WSClient) extends Controller with PanDomainAuthActions {
+
+  def previewCapiProxy(path: String) = APIAuthAction { request =>
+    val httpClient = new OkHttpClient()
```
LATaylor-guardian: Are you sure you want to create a new client each time this function is called? Really it should be a member of the class, or wrapped by an object and used as a singleton.

conf/routes
```diff
@@ -27,3 +27,7 @@ POST  /api/live/:atomType/:id                    controllers.App.publishAtom(ato
 
 PATCH /api/preview/:atomType/:id                 controllers.App.updateAtomByPath(atomType: String, id: String)
 PUT   /api/preview/:atomType/:id                 controllers.App.updateEntireAtom(atomType: String, id: String)
+
+
+#Support
+GET /support/previewCapi/*path                   controllers.Support.previewCapiProxy(path: String)
```
LATaylor-guardian: Why is this `*path` and not `:path`?

app/controllers/Support.scala
clloyd: Yeah good point (I stole this code from another project!) Will update

conf/routes
clloyd: So we can match /support/previewCapi/atom/cta/1234 if we wanted to (https://playframework.com/documentation/2.5.x/ScalaRouting#Dynamic-parts-spanning-several-/)

build.sbt
```diff
@@ -17,7 +17,8 @@ libraryDependencies ++= Seq(
   "com.gu"                   %  "kinesis-logback-appender"    % "1.3.0",
   "com.gu"                   %% "pan-domain-auth-play_2-5"    % "0.4.1",
   "io.circe"                 %% "circe-parser"                % "0.7.0",
-  "net.logstash.logback"     %  "logstash-logback-encoder"    % "4.2"
+  "net.logstash.logback"     %  "logstash-logback-encoder"    % "4.2",
+  "com.squareup.okhttp"      % "okhttp"                       % "2.4.0"
```
philmcmahon: What's the motivation for using okhttp rather than just Play WS?

clloyd: So the main reason is because it's been used in TagManager, Media Atom Maker etc the other projects, but I can switch it out

public/js/components/AtomEdit/AtomEdit.js
```diff
@@ -2,6 +2,8 @@ import React, { PropTypes } from 'react';
 import {CTAEditor} from './CustomEditors/CTAEditor';
 import {RecipeEditor} from './CustomEditors/RecipeEditor';
 import {ExplainerEditor} from './CustomEditors/ExplainerEditor';
+import {StoryQuestionsEditor} from './CustomEditors/StoryQuestionsEditor';
```
shaundillon: As these are the only exported things inside their files, is there much point in using `{Module}` over making them the default export?

public/js/components/FormFields/FormFieldTagPicker.js
```diff
@@ -0,0 +1,109 @@
...
+    fieldName: PropTypes.string,
+    fieldValue: PropTypes.string,
+    fieldPlaceholder: PropTypes.string,
+    fieldErrors: PropTypes.arrayOf(errorPropType),
+    formRowClass: PropTypes.string,
+    onUpdateField: PropTypes.func
+  };
+
+  state = {
+    suggestions: undefined,
```
shaundillon: Is it better to say `undefined` of `null`? As `null` is more specifically, a blank object reference. You'd till get the advantage of falsyness, but match the type.

public/js/components/FormFields/FormFieldTagPicker.js
```diff
@@ -0,0 +1,109 @@
...
+    this.props.onUpdateField(id);
+
+    this.setState({
+      suggestions: undefined,
+      searchText: ""
+    });
+  }
+
+  resetTag = () => {
+    this.props.onUpdateField(undefined);
```
shaundillon: Can't this just be sent with no params? The param on the other side would be `undefined` anyway

public/js/services/capi.js
```diff
@@ -0,0 +1,11 @@
+export function searchTags(seachText) {
```
shaundillon: Misspelling of searchText

public/js/services/capi.js
```diff
@@ -0,0 +1,11 @@
...
+  return fetch(
+    `/support/previewCapi/tags?q=${seachText}`,
```
shaundillon: Also misspelled here, so at least it works!

shaundillon:
Mine are mostly just questions, good work!

public/js/components/AtomEdit/AtomEdit.js
clloyd: When we're less pushed I think we should move to using the index.js file in the folder trick like ManagedForm

public/js/components/FormFields/FormFieldTagPicker.js
clloyd: Ok! Will switch over

public/js/components/FormFields/FormFieldTagPicker.js
clloyd: I just thought it was clearer? this is explicitly sending `null` through

public/js/services/capi.js
clloyd: I'm consistently incorrect

clloyd:
@philmcmahon @shaundillon This ok to merge?

philmcmahon:
👍 🐟

shaundillon:
@clloyd Yup!

PR #75, opened 2017-02-28
Move validation store to form state

shaundillon:


public/js/components/ManagedEditor/ManagedEditorForm.js
```diff
@@ -1,19 +1,46 @@
...
+  };
+
+  state: {
+    id: '',
+    formErrors: {}
+  }
+
+  componentWillMount() {
+    this.setState({
+      id: _uniqueId('form')
```
clloyd: Lodash to the rescue!

PR #77, opened 2017-03-01
fix https regex and jump over errors when validation passes

shaundillon:


public/js/util/validateField.js
```diff
@@ -5,7 +5,7 @@ const validateField = (fieldValue, isRequired: false, customValidation) => {
   const errors = [];
 
   const checkError = (err) => {
-    if (!(err instanceof FieldError)) {
+    if (!(err instanceof FieldError) && err !== true) {
```
philmcmahon: Maybe a comment here would be good to explain how an error can be a boolean?

public/js/util/validators.js
```diff
@@ -8,7 +8,7 @@ import FieldError from '../constants/FieldError';
 
 export const isHttpsUrl = (value) => {
   const stringValue = typeof value === 'string' ? value : '';
-  if (stringValue.match(/^(https?:\/\/)?([\da-z\.-]+)\.([a-z\.]{2,6})([\/\w \.-]*)*\/?$/)) {
+  if (stringValue.match(/^(https:\/\/)?([\da-z\.-]+)\.([a-z\.]{2,6})([\/\w \.-]*)*\/?$/)) {
```
clloyd: Hahahaha oops

PR #81, opened 2017-03-02
Atom Stats

shaundillon:
BONUS - Hitting capi with the usage URLs to get more details

Possible improvements: Add composer and viewer URLs to config so we can link out to them in the usages.

picture 117


public/js/components/CAPIQueryBuilder/CapiQueryBuilder.js
```diff
@@ -0,0 +1,10 @@
+// import React from 'react';
```
philmcmahon: Probably best to delete this - just copy pasted code!

shaundillon:
Closed as it's now included in #82

PR #82, opened 2017-03-03
Add atom finder

philmcmahon:
This PR contains all the changes in https://github.com/guardian/atom-workshop/pull/81.

Config update needed!


public/js/components/AtomList/AtomList.js
```diff
@@ -0,0 +1,173 @@
...
+    getUserString = (changeRecord) => {
+        if (!changeRecord || !changeRecord.user) return 'unknown';
+        if (!changeRecord.user.firstName || !changeRecord.user.lastName) return changeRecord.user.email;
+        return `${changeRecord.user.firstName} ${changeRecord.user.lastName}`;
+    };
+
+    getDateString = (changeRecord) => {
+        if (!changeRecord) return 'unknown';
+        const date = new Date(changeRecord.date);
+        return `${date.toLocaleDateString()} ${date.toLocaleTimeString('en-GB', {formatMatcher: "basic"})}`;
```
clloyd: We have date-fns pulled into this project now, so that could be an option if you want to make this code a bit nicer?

public/js/components/AtomList/AtomList.js
```diff
@@ -0,0 +1,173 @@
...
+                                <th className="atom-list__header">Title</th>
+                                <th className="atom-list__header">Type</th>
+                                <th className="atom-list__header">Last Modified</th>
+                                <th className="atom-list__header">Created by</th>
+                                <th className="atom-list__header">Publication status</th>
+                                <th className="atom-list__header"> </th>
+                            </tr>
+                        </thead>
+                        <tbody>
+                        {this.props.atomList.map((atom) => (
```
clloyd: Just for clarity... can this be pulled out into a seperate render method? e.g. `this.props.atomList.map(renderListItem)`

clloyd: I imagine that this will could be pulled out into a seperate component entirely when the complexity increases

public/js/components/FormFields/FormFieldTagPicker.js
```diff
@@ -1,7 +1,7 @@
 import React, {PropTypes} from 'react';
 import ShowErrors from '../Utilities/ShowErrors';
 import { errorPropType } from '../../constants/errorPropType';
-import {searchTags} from '../../services/capi';
+import capi from '../../services/capi';
```
clloyd: Do we need to import the whole of capi?

public/js/components/Header/Header.js
```diff
@@ -82,6 +82,19 @@ class Header extends React.Component {
     return false;
   }
 
+  renderCreateNewButton = () => {
+    if(location.pathname === "/list") {
```
clloyd: Can we not pull this from `location`? There should be a `this.props.router` with a route on it. That's the actual source of truth with what route we're currently on (not the URL bar...)

public/js/reducers/atomListReducer.js
```diff
@@ -0,0 +1,12 @@
+
+
+export default function atomList(state = null, action) {
+    switch (action.type) {
+
+        case 'ATOM_LIST_GET_RECEIVE':
+            return action.atomList || false;
```
clloyd: Would it make sense to default to `[]` at the start and on recieve rather than false/null, the propTypes check is looking for an array only

public/js/services/capi.js
```diff
@@ -1,11 +1,64 @@
...
-      credentials: 'same-origin'
-    }
-  )
-  .then((res) => res.json())
-  .then((json) => Promise.resolve(json.response.results));
-}
+import { pandaFetch } from './pandaFetch';
+import { uriEncodeParams } from '../util/uriEncodeParams';
+
+export default {
```
clloyd: How come we're exporting an object and not individual functions?

public/js/components/AtomList/AtomList.js
shaundillon: @philmcmahon You can see an example of this in AtomStats.js

public/js/components/FormFields/FormFieldTagPicker.js
shaundillon: @clloyd - I agree but isn't this how it's done with the VideosApi?

public/js/services/capi.js
shaundillon: makes sense. If this is the case, it'll need doing in VideosApi too

public/js/components/AtomListItem/AtomListItem.js
```diff
@@ -0,0 +1,74 @@
...
+    renderEditorLink = (atom) => {
+        const title = atomTitleExtractor(atom);
+        if (supportedAtomTypes.map((t)=>t.type).indexOf(atom.atomType) !== -1) {
+            return <Link to={`/atoms/${atom.atomType}/${atom.id}/edit`}
+                         className="atom-list__link atom-list__editor-link"
+                         key={atom.id}>
+                {title}</Link>;
+        }
+        function externalEditorUrl(editorUrls) {
+            switch (atom.atomType) {
```
shaundillon: if this gets any larger I think we'll have to split it out into a util. It's fine for now, though

shaundillon:
The list stull all looks good to me 👍

PR #83, opened 2017-03-06
Re order routes file to match controller order in AppComponents.

philmcmahon:
When I try to run master at the moment I get a compile error. This fixes that, though if nobody else has the error then maybe everything is fine!


shaundillon:
I had it too briefly 👍

PR #89, opened 2017-03-15
Make header more modular, hide editor buttons when not in editor

philmcmahon:
This fixes a bug where editor buttons still show after clicking on an atom and then going back. It also restructures how the header is laid out in a (hopefully) more modular fashion.


public/js/components/Header/Header.js
```diff
@@ -39,6 +39,13 @@ class Header extends React.Component {
     this.props.atomActions.takeDownAtom(this.props.atom);
   }
 
+  isEditor = () => {
+    return location.pathname.search(new RegExp('\/atoms\/.*\/.*\/edit', 'g')) >= 0;
```
shaundillon: This would be nicer if we used react-router's RouteComponent stuff - https://reacttraining.com/react-router/web/example/url-params

public/js/components/Header/Header.js
```diff
@@ -55,62 +62,48 @@ class Header extends React.Component {
...
-      );
-    }
-    return false;
+    );
   }
 
-  renderTakeDownButton = (atomPublishState) => {
-
-    if(atomPublishState.id !== 'draft') {
+  renderHeaderLeft = () => {
```
shaundillon: Not sure I like the name, we may want to use this elsewhere. Sorry for being finicky!

public/js/components/Header/Header.js
```diff
@@ -55,62 +62,48 @@ class Header extends React.Component {
...
 
-      return (
+    return (
         <div className="toolbar__container">
-          <div className="toolbar__item">
-            {this.props.presence ? <PresenceIndicator presence={this.props.presence} /> : false}
-          </div>
-          <button disabled={atomPublishState.id === 'published'} type="button" onClick={this.publishAtom} className="toolbar__item toolbar__button">Publish</button>
-          {this.renderTakeDownButton(atomPublishState)}
+          {this.props.presence ? <PresenceIndicator presence={this.props.presence} /> : false}
```
shaundillon: I know it was me, but we may have the same problem here as presence will still exist in state after visiting one page.

public/js/components/Header/Header.js
philmcmahon: I don't think we can do this as Header is a child of Page, which just has the route `/` - so the id and atomType parameters aren't available?

PR #91, opened 2017-03-15
Reindexing enpoints

marialivia16:


app/db/AtomDataStores.scala
```diff
@@ -1,75 +1,24 @@
 package db
```
philmcmahon: So much wonderful red!!!! 🎆

PR #93, opened 2017-03-15
Add units to ingredients

shaundillon:
picture 131

### DIFF
--- a/public/js/components/AtomEdit/CustomEditors/RecipeFields/Ingredient.js
+++ b/public/js/components/AtomEdit/CustomEditors/RecipeFields/Ingredient.js
@@ -157,10 +157,10 @@ class Quantity extends React.Component {
 }
 
 class Units extends React.Component {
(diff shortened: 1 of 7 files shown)


public/js/components/AtomEdit/CustomEditors/RecipeFields/Ingredient.js
```diff
@@ -141,3 +155,26 @@ class Quantity extends React.Component {
...
+    updateUnit: PropTypes.func.isRequired
+  };
+
+  updateIngredientUnit = (value) => {
+    this.props.updateUnit(value);
+  }
+
+  render () {
+    return (
+      <FormFieldSelectBox
```
philmcmahon: There are quite a lot of different unit types. Might it be possible to use something similar to the tag picker component to allow people to filter which unit they want?

philmcmahon:
👍

PR #94, opened 2017-03-16
MAKE RECIPES GREAT AGAIN

shaundillon:
A few assumed UX tweaks for the recipe atom maker in particular, but can be used across the board.
*Warning, there's some gross CSS in here*
picture 133
picture 134


philmcmahon:
I don't know enough to have a strong opinion on how gross the css here is, but looks good to me!! :cucumber:

PR #96, opened 2017-03-17
Disable presence client side when disabled via config switch.

philmcmahon:
This gets rid of some console errors when running locally.


shaundillon:
👍

PR #97, opened 2017-03-17
Add fields to cta editor to match cta maker

philmcmahon:
The background text field was incorrectly named so I've renamed that. We were also missing the campaign field, which is stored in the top level labels array. In CTA maker it's not an array input field - just a single item, but I'm hoping its ok to do it as an array here!

CTA Maker:
https://cta-atom-maker.code.dev-gutools.co.uk/#/atom/96e3af8c-87e6-4706-87a9-8a29e82361ee
CAPI representation:
http://content.code.dev-guardianapis.com/atom/cta/96e3af8c-87e6-4706-87a9-8a29e82361ee

screen shot 2017-03-17 at 10 59 53


shaundillon:
Yep, you've used everything correctly. 👍

PR #99, opened 2017-03-20
Targeting Integration

clloyd:
This adds targeting integration, as part of implementation I determined that `unfetch` didn't polyfill enough features around cors/creds so this has been reverted back to the full polyfill

This also adds a general date select `FormFieldDateInput` which leverages [react-dates](https://github.com/airbnb/react-dates) which unfortunately has a `moment` dependency :( but is a really nice date picker.

While working with the form validation I refactored/fixed a few aspects of the existing form validation
 - Forms are now validated on mounting
 - Forms will only show an error after the first update is triggered
 - `ManagedForm` now exposes a function which will supply the latest errors Object
 - `ManagedForm` also exposes a function which will send a `true`/`false` for the form validation state... much easier to disable a submit button

screen shot 2017-03-20 at 14 06 24
screen shot 2017-03-20 at 14 06 30
screen shot 2017-03-20 at 14 07 01
screen shot 2017-03-20 at 14 07 10


public/js/components/ManagedEditor/ManagedEditorForm.js
```diff
@@ -23,12 +25,23 @@ export class ManagedForm extends React.Component {
...
+    });
+
     this.setState({
-      formErrors: {
-        [fieldName]: fieldErrors
-      }
+      formErrors: formErrors
     });
+
+    this.props.onFormErrorsUpdate && this.props.onFormErrorsUpdate(formErrors);
```
shaundillon: What are these doing?

clloyd: The parent can pass down a function that is called with the latest errors (similar to an onUpdate) this checks if it's been passed and calls it with the latest errors

shaundillon:
Aside from my comment 👍  Nice work on the validation!

PR #100, opened 2017-03-21
handle external atom links in their own page

shaundillon:
picture 137


clloyd:
I think this needs the create button and maybe make the design a bit more similar to the AtomEditor? with top card etc

shaundillon:
@clloyd - I'll change up the design, but I'm not sure about the create new button. it may lead people to believe they're creating a new one of that particular atom type?

PR #104, opened 2017-03-23
Atom search styling

shaundillon:
Before
picture 149

After:
picture 148

This includes some new fields, they're essentially simpler versions of their standard ones. This was far easier than making all of the classes dynamic on the existing fields.


philmcmahon:
Looks great! 👍

PR #106, opened 2017-03-27
Recipe tags

shaundillon:
Added recipe tags list from Recipeasy and styled them out.

picture 153


shaundillon:
Added do not merge for now while I sort out the horrendous UI

public/js/constants/recipeData.js
```diff
@@ -0,0 +1,737 @@
...
+  'Nut free',
+  'Gluten free',
+  'Dairy free',
+  'Egg free',
+  'Vegetarian',
+  'Vegan'
+];
+
+export const chefs = [
+  'Nigel Slater and Observer readers',
```
philmcmahon: We should think about whether we want to link these up to contributor tags - e.g. Nigel Slater has a contrbutor tag: https://tagmanager.gutools.co.uk/tag/16196

philmcmahon:
This looks good to me! Might be worth us thinking about how chefs/contributor tags come together at some point (if at all) but not worth bothering about now. 👍

public/js/constants/recipeData.js
shaundillon: I agree, but was just matching the functionality of recipeasy for now. Adding contributors would be a great improvement though

PR #107, opened 2017-03-29
Move form errors to redux store

shaundillon:
This is based on [Reeta's PR](https://github.com/guardian/media-atom-maker/pull/344/files) in the Media atom maker. Moving our form errors state out of the form scope and into the redux store,
in an object sorted by a form title.

/cc @clloyd


public/js/reducers/formErrorsReducer.js
```diff
@@ -0,0 +1,16 @@
+export default function formErrors(state = {}, action) {
+
+  let formName, newFormErrors, currentFormErrors, updatedFormErrors, updatedForm;
```
clloyd: let?

shaundillon: I just copied these from Reeta's haha. Will change.

clloyd: I was more meaning `const`?

clloyd:
@shaundillon I was referring to making them immutable, vars/lets in reducers make me a bit  nervous

PR #108, opened 2017-03-30
Reflect search in query strings for easier navigation

shaundillon:
this turned out a little more complicated than anticipated so here's a brief explanation:

*any* component can store its own `queryParams` in state using the usual binding action creators/calling them workflow.

There's some custom middleware (which currently only checks for a specific action, but can be extended) and converts this object into a string, which is pushed to the browser history. 

On startup, there's a function which reads the URL params into `queryParams` in state, so that people can navigate to a URL with params and see the same as anyone else.

## Possible improvement

- instead of just pushing the URL into `browserHistory` actually change the action in flight so that the call to capi doesn't need to do it's own encoding. (The problem here is, capi doesn't like the formatting of Arrays I've used).

/cc @clloyd


public/js/components/AtomList/AtomList.js
```diff
@@ -27,17 +37,46 @@ class AtomList extends React.Component {
...
+
+        const convertUriListToArray = (uriList) => {
+          if (uriList.length === 0) {
+            return false;
+          }
+
+          return uriList.split(',');
+        };
+
+        const newParams = Object.assign({}, this.state.params, {
```
clloyd: Do you need to de-url-encode this? will this result in the url encoded versions being re-url-encoded

shaundillon: yeah I agree, it was only to get around disrupting what was already in place. I'll just adjust the capi method to take an encoded string.

LATaylor-guardian:
@shaundillon Great, nice work!

PR #109, opened 2017-03-31
Add ability to pass custom stats link

clloyd:
This will replace the /stats link on the dashboard with a custom link (much like custom create url) 

cc @shaundillon @LATaylor-guardian


LATaylor-guardian:
Thanks for this 👍

PR #117, opened 2017-04-13
Make publish button green, takedown button red.

philmcmahon:
Not sure the css padding stuff I've written here is the nicest way of doing things, let me know!

Currently looks like this:

screen shot 2017-04-13 at 14 26 55
screen shot 2017-04-13 at 14 27 13

### DIFF
--- a/public/js/components/Header/EditHeader.js
+++ b/public/js/components/Header/EditHeader.js
@@ -66,7 +66,9 @@ class EditHeader extends React.Component {
   renderTakeDownButton = (atomPublishState) => {
     if(atomPublishState.id !== 'draft') {
       return (
-        <button type="button" disabled={atomPublishState.id === 'taken-down'} onClick={this.takeDownAtom} className="toolbar__item toolbar__button">Take down</button>
+          <div className="bold-button-padright">
+        <button type="button" disabled={atomPublishState.id === 'taken-down'} onClick={this.takeDownAtom} className="btn btn--red">Take down</button>
+          </div>
       );
     }
     return false;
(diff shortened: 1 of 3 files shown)


public/js/components/Header/EditHeader.js
```diff
@@ -66,7 +66,9 @@ class EditHeader extends React.Component {
   renderTakeDownButton = (atomPublishState) => {
     if(atomPublishState.id !== 'draft') {
       return (
-        <button type="button" disabled={atomPublishState.id === 'taken-down'} onClick={this.takeDownAtom} className="toolbar__item toolbar__button">Take down</button>
+          <div className="bold-button-padright">
```
shaundillon: As we (kind of) use the BEM naming convention, the padding on the right of this should be a "modifier" of the original class. So this one should have the classes `bold-button bold-button--padright` where the second has the padding but the first keeps all the same styles.

You _could_ even take advantage of pseudoselectors instead, so in your css doing `.bold-button:not(:last-child) { padding-right: 5px; }` for example.

Also, could the button not push itself away from the sides using a margin instead of having a containing div? I don't know that this is possible as they're all flex items and a bit funny.

philmcmahon: Thanks, updated!

public/styles/components/_buttons.scss
```diff
@@ -23,6 +23,15 @@ button {
     background: $brandColor;
   }
 
+  &--margin {
+    margin: 5px;
+    &:not(:last-child) {
+      margin-right: 0px;
```
shaundillon: I think these need to be the other way round, right?

philmcmahon:
Prematurely merged this! But the final thing looked like this:
 &--margin {
+    margin-top: 5px;
+    margin-bottom: 5px;
+    margin-right: 5px;
+    &:first-child {
+      margin-left: 5px;
+    }
+  }

PR #120, opened 2017-04-24
Tweak queryParams for edge cases on search

shaundillon:
The searchParams were being overwritten with an empty object. This merges that empty object with the existing one to stop errors on the managed search fields.

Bonus grammar tweaks and cleaner initialisation of the routerMiddleware.


philmcmahon:
👍

PR #121, opened 2017-04-27
Some UX tweaks for story questions

rcrphillips:
Nathan's been using the story question tool a bit more and has come across a few UX issues. @shaundillon and I..... mainly @shaundillon, have removed a couple of unnecessary fields and ensured that there's a a random vale set for the capi model.


public/js/components/AtomEmbed/CreateTargetForm.js
```diff
@@ -67,9 +67,6 @@ class CreateTargetForm extends React.Component {
           <ManagedField fieldLocation="title" name="Title" isRequired={true}>
             <FormFieldTextInput/>
           </ManagedField>
-          <ManagedField fieldLocation="description" name="Description" isRequired={true}>
```
shaundillon: Just had a look at the targeting tool and it looks like this is optional, which is good. @currysoup , am I correct?

PR #122, opened 2017-05-05
Storyquestions default HTML snippet

regiskuckaertz:
This is in conjunction with the work done [here](https://github.com/guardian/flexible-content/pull/2821) to preview a story questions atom in the suggestions.

I expect this is not as idiomatic as this project's authors would like it to be, so please do comment.

One question I will answer preemptively is that I followed the thread of `CreateAtomFields.defaultHtml` and it seems to be coming from [some React component](https://github.com/guardian/atom-workshop/blob/master/public/js/actions/AtomActions/createAtom.js#L32), and so I didn't want to define the default HTML there because it felt really fragile.

### DIFF
--- a/app/util/AtomElementBuilders.scala
+++ b/app/util/AtomElementBuilders.scala
@@ -51,7 +51,7 @@ object AtomElementBuilders {
   }
 
   def buildDefaultHtml(atomType: AtomType, atomData: AtomData): String = {
-    s"""<div class="atom-${atomType.name}">${buildHtml(atomType, atomData)}</div>"""
+    s"""<div class="atom-${atomType.name}">${buildHtml(atomType, atomData).getOrElse("")}</div>"""
   }
 
   def buildHtml(atomType: AtomType, atomData: AtomData): Option[String] = atomType match {
--- a/app/util/AtomElementBuilders.scala
+++ b/app/util/AtomElementBuilders.scala
@@ -44,29 +44,25 @@ object AtomElementBuilders {
       title = createAtomFields.flatMap(_.title),
       id = java.util.UUID.randomUUID.toString,
       atomType = atomType,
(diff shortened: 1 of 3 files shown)


app/util/AtomElementBuilders.scala
```diff
@@ -44,14 +44,29 @@ object AtomElementBuilders {
...
+  }
+
+  def buildStoryQuestionsHtml(sqData: AtomData): String = {
+    val questions: Option[String] = for {
+      eqs <- sqData.asInstanceOf[AtomData.Storyquestions].storyquestions.editorialQuestions
+    } yield {
+      eqs flatMap (_.questions map { q => s"<li>${q.questionText}</li>" }) mkString ""
+    }
+
+    questions map ("<ul>" ++ _ ++ "</ul>") getOrElse ""
```
akash1810: should we return `<div class="atom-${atomType.name}"></div>` rather than empty string in the unhappy path for consistency?

marialivia16: Out of curiosity: I understand you want to construct a list with the questions, but you seem to not wrap it in a `div` and I was wondering why.
I mean having something like:
```
<div class="atom-${atomType.name}"><ul>..</ul>...</div>
```

regiskuckaertz: Oh, that's a very good point. Indeed I could combine that with your comment above to avoid having to repeat the default HTML snippet 👍

regiskuckaertz: Yes, it is sensible (see below)

regiskuckaertz:
😮  yoohoo! Appreciate your help guys, and super happy this got through so fast 💨

regiskuckaertz:
Side question: is `CreateAtomFields.defaultHtml` even necessary?

My initial thinking was that the HTML generation could be done via an ADT, but `CreateAtomFields` lacks the necessary `atomData` and `atomType` values. Food for thought 😃

PR #135, opened 2017-06-27
Remove title field from qanda editor

tomrf1:
The atom-level `title` field will be used (it is mandatory when creating a new atom).
Having this extra field is just confusing.
Also renamed "Body" label to "Answer"

picture 10


regiskuckaertz:
😂

regiskuckaertz:
I left an optional title in the model, but if it's not useful I can remove it

tomrf1:
@regiskuckaertz it's not really a problem having it in the model

PR #142, opened 2017-07-11
Build default HTML for snippets

regiskuckaertz:
Some part of me thinks this work belongs in Composer, because that is the only place it is every going to be used. Meanwhile, all CAPI requests will carry this additional payload even if they don't need it.


regiskuckaertz:
Once again this builds correctly on my machine but times out in CI 😩

app/util/AtomElementBuilders.scala
```diff
@@ -63,14 +64,49 @@ object AtomElementBuilders {
   }
 
   def buildHtml(atomType: AtomType, atomData: AtomData): Option[String] = atomType match {
-    case AtomType.Storyquestions => buildStoryQuestionsHtml(atomData)
+    case AtomType.Storyquestions => buildStoryQuestionsHtml(atomData.asInstanceOf[AtomData.Storyquestions].storyquestions)
```
tomrf1: Not really an issue, but you can avoid using `asInstanceOf` by matching on the atomData, e.g. https://github.com/guardian/story-manager/blob/master/app/models/Atom.scala#L12

regiskuckaertz: Brilliant! I did that.

regiskuckaertz:
Ping. Is everybody happy?

tomrf1:
👍 I'm happy with the code, and using this as a quick fix.
But I agree with your comment that this belongs in composer, not capi.

regiskuckaertz:
Thanks! Yep, I'll card this in Trello and will move that to composer in short time.

PR #143, opened 2017-07-11
Separate snippets section on create page

tomrf1:
picture 6


annebyrne:
The code looks fine, but I think we need to decide on the groupings and how this is conveyed to Editorial as per Slack conversation

tomrf1:
I've added a blurb

annebyrne:
:+1:

PR #144, opened 2017-07-11
Support formatting in qanda answer

tomrf1:
This uses the existing `FormFieldsScribeEditor` component (which isn't currently being used). I've modified it to optionally hide the toolbar, since all we want is the ability to add `<p>` tags when the user creates a newline.
The suggested word count is 150 - it reports a warning if you go over, but still allows it.

(this will need a frontend change to use the body as html)

picture 9

### DIFF
--- a/public/js/components/AtomEdit/CustomEditors/QAndAFields/QAItem.js
+++ b/public/js/components/AtomEdit/CustomEditors/QAndAFields/QAItem.js
@@ -1,6 +1,6 @@
 import React, { PropTypes } from 'react';
 import {ManagedForm, ManagedField} from '../../../ManagedEditor';
-import FormFieldTextArea from '../../../FormFields/FormFieldTextArea';
+import FormFieldsScribeEditor from '../../../FormFields/FormFieldScribeEditor';
 
 export class QAItem extends React.Component {
   static propTypes = {
@@ -29,7 +29,7 @@ export class QAItem extends React.Component {
       <div className="form__field">
         <ManagedForm data={value} updateData={this.updateItem} onFormErrorsUpdate={this.props.onFormErrorsUpdate} formName="qaEditor">
           <ManagedField fieldLocation="body" name="Answer" isRequired={true}>
-            <FormFieldTextArea/>
+            <FormFieldsScribeEditor showWordCount={true} suggestedLength={150} showToolbar={false}/>
           </ManagedField>
         </ManagedForm>
       </div>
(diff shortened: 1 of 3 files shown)


public/js/components/AtomEdit/CustomEditors/QAndAFields/QAItem.js
```diff
@@ -1,6 +1,6 @@
 import React, { PropTypes } from 'react';
 import {ManagedForm, ManagedField} from '../../../ManagedEditor';
-import FormFieldTextArea from '../../../FormFields/FormFieldTextArea';
+import FormFieldsScribeEditor from '../../../FormFields/FormFieldScribeEditor';
```
regiskuckaertz: Oh, I did not know this one existed 👍

public/js/components/FormFields/FormFieldScribeEditor.js
```diff
@@ -70,30 +72,32 @@ export class ScribeEditor extends React.Component {
...
-        ol: {},
-        li: {}
-      }
-    }));
+    if (this.props.showToolbar !== false) {
+      this.scribe.use(scribePluginToolbar(this.refs.toolbar));
+
+      // Configure Scribe plugins
+      this.scribe.use(scribePluginLinkPromptCommand());
+      this.scribe.use(scribeKeyboardShortcutsPlugin({
```
regiskuckaertz: Not sure we don't need this one and the following plugins. Would people want to emphasise words or add links at some point?

public/js/components/FormFields/FormFieldScribeEditor.js
```diff
@@ -70,30 +72,32 @@ export class ScribeEditor extends React.Component {
...
+        bold: function (event) { return event.metaKey && event.keyCode === 66; }, // b
+        italic: function (event) { return event.metaKey && event.keyCode === 73; }, // i
+        linkPrompt: function (event) { return event.metaKey && !event.shiftKey && event.keyCode === 75; }, // k
+        unlink: function (event) { return event.metaKey && event.shiftKey && event.keyCode === 75; } // shft + k
+      }));
+
+      this.scribe.use(scribePluginSanitizer({
```
regiskuckaertz: I think this plugin is needed if people are copy-pasting content from somewhere like Google Docs. There's a lot of useless markup that often comes with such.

regiskuckaertz:
giphy

PR #147, opened 2017-07-28
Storyquestions closeDate field

tomrf1:
Currently Editorial have to remove a storyquestions atom from all content when they want to stop displaying the atom.
Checking this box will tell the platforms not to render the atom.
The field is a dateTime, and checking this box sets it to the current time (UTC).

In future I'd like to make this a `Close` button in a sidebar on the right, using the 'custom actions' design being developed in this branch - https://github.com/guardian/atom-workshop/compare/rk-sq-notif.

But for now...
picture 22

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ version := "1.0"
 scalaVersion := "2.11.8"
 
 lazy val awsVersion = "1.11.8"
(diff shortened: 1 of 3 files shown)


mchv:
👍

PR #149, opened 2017-08-08
Suggested content

tomrf1:
In the `stats` page we currently only display a list of content that uses the atom.
This change adds a list of suggested content. This list is based on any target tags. Only content from the last 7 days is included.

E.g. before adding to content:
picture 7
And after:
picture 8

### DIFF
--- a/public/js/actions/AtomActions/getSuggestedContent.js
+++ b/public/js/actions/AtomActions/getSuggestedContent.js
@@ -1,0 +1,72 @@
+import {getContentByTags} from '../../services/capi';
+import {fetchTargetsForAtomPath} from '../../services/TargetingApi';
+import {logError} from '../../util/logger';
+
+
+function requestSuggestedContent(atomType, id) {
+  return {
+    type:       'SUGGESTED_CONTENT_GET_REQUEST',
+    atomType:   atomType,
+    id:         id,
+    receivedAt: Date.now()
+  };
+}
+
+function receiveSuggestedContent(suggestedContent) {
+  return {
+    type:       'SUGGESTED_CONTENT_GET_RECEIVE',
+    suggestedContent: suggestedContent,
+    receivedAt: Date.now()
+  };
+}
+
+function errorReceivingSuggestedContent(error) {
+  logError(error);
+  return {
+    type:       'SHOW_ERROR',
+    message:    'Could not get atom usages',
+    error:      error,
+    receivedAt: Date.now()
+  };
+}
+
+/**
+ * Gets tags from Target api.
+ * Queries capi for content with those tags, filtering out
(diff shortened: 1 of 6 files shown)


public/js/actions/AtomActions/getSuggestedContent.js
```diff
@@ -0,0 +1,74 @@
...
+ * Gets tags from Target api.
+ * Queries capi for content with those tags, filtering out
+ * any content which already have this atom.
+ */
+export function getSuggestedContent(atomId, atomType) {
+  const path = `/atom/${atomType.toLowerCase()}/${atomId}`;
+  const pluralType = `${atomType}s`;
+
+  return dispatch => {
+    dispatch(requestSuggestedContent());
```
regiskuckaertz: shouldn't you pass `atomId` and `atomType` here?

public/js/actions/AtomActions/getSuggestedContent.js
```diff
@@ -0,0 +1,74 @@
...
+    return fetchTargetsForAtomPath(path)
+      .then((targets) => {
+        return targets.map((target) => {
+          return target.tagPaths;
+        });
+      })
+      .then(tags => {
```
regiskuckaertz: you can spare one level of nesting this way:
```
.then(tags => getContentByTags(tags, pluralType))
.then(content => {
    ...
})
.catch(error => {
    dispatch(errorReceivingSuggestedContent(error));
});
```

it won't change the world, I know 😅

public/js/actions/AtomActions/getSuggestedContent.js
```diff
@@ -0,0 +1,74 @@
...
+                const atomIdx = c.atoms[pluralType].find(a => {
+                  return a.id === atomId;
+                });
+                return atomIdx === -1;
+              } else {
+                return true;
+              }
+            });
+
+            suggested.sort((first, second) => new Date(second.fields.creationDate) - new Date(first.fields.creationDate));
```
regiskuckaertz: Isn't `creationDate` a number?

public/js/services/capi.js
```diff
@@ -67,3 +67,18 @@ export const getByPath = (path) => {
     return Promise.resolve(json.response.content);
   });
 };
+
+export const getContentByTags = (tags, atomType) => {
+  const date = new Date(new Date().setDate(new Date().getDate()-7));
```
regiskuckaertz: 😮

public/js/actions/AtomActions/getSuggestedContent.js
tomrf1: It's a datetime string - content.guardianapis.com/search?show-fields=creationDate

PR #150, opened 2017-08-09
Allow image removal

tomrf1:
New red button:
picture 11

### DIFF
--- a/public/js/components/FormFields/__snapshots__/FormFieldImageSelect.spec.js.snap
+++ b/public/js/components/FormFields/__snapshots__/FormFieldImageSelect.spec.js.snap
@@ -58,10 +58,16 @@ exports[`test Should render with image when provided 1`] = `
           1215
         </div>
         <button
-          className="image-select__button"
+          className="image-select__button image-select__replace-button"
           onClick={[Function]}
           type="button">
           Replace Image
+        </button>
+        <button
+          className="image-select__remove-button btn--red"
+          onClick={[Function]}
+          type="button">
+          Remove Image
         </button>
       </div>
     </div>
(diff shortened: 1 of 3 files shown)


public/js/components/FormFields/__snapshots__/FormFieldImageSelect.spec.js.snap
```diff
@@ -58,11 +58,17 @@ exports[`test Should render with image when provided 1`] = `
...
         </div>
         <button
-          className="image-select__button"
+          className="image-select__button image-select__replace-button"
           onClick={[Function]}
           type="button">
           Replace Image
         </button>
+        <button
+          className="image-select__remove-button btn btn--red"
```
annebyrne: dumb question - why does remove have a `btn` class but replace doesn't? is it needed?

public/styles/components/form-fields/_image-select.scss
```diff
@@ -43,6 +43,15 @@
     }
   }
 
+  &__replace-button {
+    margin-right: 20px;
```
regiskuckaertz: that's one way to do it. But if there is no button on the right, you're left with a right margin that might break your layout. What you can do instead is have a left margin on any button that follows another button:
```
button + button {
  margin-left: 20px;
}
```

annebyrne:
👍  apart from my comment which isn't really an issue

tomrf1:
@regiskuckaertz it'll pass null as the value for this `ManagedField`, which means the field becomes unset in the model. I can't see a nicer way of doing this...

public/js/components/FormFields/__snapshots__/FormFieldImageSelect.spec.js.snap
tomrf1: The two generally go together, though actually the only extra styling it's getting from `btn` is the `color` because I'm overriding everything else.

PR #153, opened 2017-08-25
Improve add & delete buttons in atom editor

annebyrne:
This is probably some real ~nasty~ _hacky_ CSS so comments v welcome!! Given this makes changes to shared components I've included screenshots so you can see effect across atom editors.

Screenshots of new styling:

**Story Questions:**

screen shot 2017-08-25 at 15 33 36

**Quick Guide:**

screen shot 2017-08-25 at 15 31 38

**CTA:**

screen shot 2017-08-25 at 15 34 15

**Recipe:**

screen shot 2017-08-25 at 15 34 53

**Profile:**

screen shot 2017-08-25 at 15 35 19

**Timeline:**

screen shot 2017-08-25 at 15 35 45

### DIFF
--- a/public/js/components/FormFields/FormFieldArrayWrapper.js
+++ b/public/js/components/FormFields/FormFieldArrayWrapper.js
@@ -19,8 +19,8 @@ export default class FormFieldArrayWrapper extends React.Component {
   }
 
   onAddClick = () => {
-    const exisitingValues = this.props.fieldValue || [];
-    this.props.onUpdateField(exisitingValues.concat([undefined]));
+    const existingValues = this.props.fieldValue || [];
+    this.props.onUpdateField(existingValues.concat([undefined]));
   }
 
   renderValue(value, i) {
@@ -53,10 +53,10 @@ export default class FormFieldArrayWrapper extends React.Component {
     });
 
     return (
-      <div className={this.props.fieldClass ? this.props.fieldClass : 'form__group'}>
+      <div className={this.props.fieldClass ? this.props.fieldClass : 'form__group form__field'}>
         {this.props.numbered ? <span className="form__field-number">{`${i + 1}. `}</span> : false }
         {hydratedChildren}
-        <button className="btn form__field-btn" type="button" onClick={removeFn.bind(this, i)}>Delete</button>
+        <button className="btn form__field-btn btn--red" type="button" onClick={removeFn.bind(this, i)}>Delete</button>
       </div>
     );
   }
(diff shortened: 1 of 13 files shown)


annebyrne:
well i did not expect those images to be that large! oops. hope it's still readable, will remove if needed

public/js/components/FormFields/FormFieldArrayWrapper.js
```diff
@@ -53,10 +53,10 @@ export default class FormFieldArrayWrapper extends React.Component {
     });
 
     return (
-      <div className={this.props.fieldClass ? this.props.fieldClass : 'form__group'}>
+      <div className={this.props.fieldClass ? this.props.fieldClass : 'form__group form__field'}>
         {this.props.numbered ? <span className="form__field-number">{`${i + 1}. `}</span> : false }
         {hydratedChildren}
-        <button className="btn form__field-btn" type="button" onClick={removeFn.bind(this, i)}>Delete</button>
+        <button className="btn form__field-btn btn--red" type="button" onClick={removeFn.bind(this, i)}>Delete</button>
```
shaundillon: I know this _seems_ picky, but `btn--danger` would be more appropriate as a name here.

public/styles/components/_atom-editor.scss
```diff
@@ -18,7 +18,7 @@
   }
 
   &__form {
-    margin-top: 20px;
+    margin: 20px 0px 40px 0px;
```
shaundillon: No need for the units on the `0` values.

public/styles/components/_forms.scss
```diff
@@ -227,6 +237,13 @@
       background: $brandColor;
     }
   }
+
+  &__add {
+    padding: 10px;
+    display: table;
```
shaundillon: You wanna be 100% certain you want `display: table` here. It means a lot of CSS properties won't work on it any more

shaundillon:
Aside from those bits, it looks like the "Public title" and "Link URL" fields are missing a class

public/styles/components/_forms.scss
annebyrne: @regiskuckaertz would know better than I the benefits / merits of this approach

annebyrne:
@shaundillon Can you be clearer on this point please?

> looks like the "Public title" and "Link URL" fields are missing a class

public/js/components/FormFields/FormFieldArrayWrapper.js
annebyrne: This is using a pre-existing class so not going to change for the moment

public/styles/components/_forms.scss
regiskuckaertz: That's correct, the `padding` for instance can be removed here. The alternative (without changing the markup) would be to make the ancestor a flex column container and then flush-right the children, but I think this may have more side-effects.

shaundillon:
@annebyrne In the screenshots, the "Public title" and "Link URL" fields have default browser borders. It appears they're missing the `form__field` class.

PR #155, opened 2017-09-13
Create Exact Target email lists for storyquestions atoms

tomrf1:
This adds a new optional "atom actions" sidebar to the right of the atom editor screen.
This is currently only used for storyquestions atoms to support the creation/deletion of an Exact Target list (currently this is done manually).

The interaction with Exact Target is done via a separate library.

### Before creating a list:
picture 5

### After creating a list:
picture 6

### DIFF
--- a/app/controllers/AtomActions.scala
+++ b/app/controllers/AtomActions.scala
@@ -1,0 +1,58 @@
+package controllers
+
+import cats.syntax.either._
+import com.gu.contentatom.thrift.{Atom, EventType}
+import db.AtomDataStores._
+import db.AtomWorkshopDBAPI
+import models._
+import play.api.libs.ws.WSClient
+import play.api.mvc.Controller
+import services.AtomPublishers._
+import services.NotificationLists
+import util.AtomLogic._
+import util.AtomUpdateOperations._
+import io.circe._
+import com.gu.fezziwig.CirceScroogeMacros._
+import com.gu.pandomainauth.action.UserRequest
+
+class AtomActions(val wsClient: WSClient, val atomWorkshopDB: AtomWorkshopDBAPI) extends Controller with PanDomainAuthActions {
+
+  def createNotificationList(atomType: String, id: String) = AuthAction { req =>
+    APIResponse {
+      for {
+        currentDraftAtom <- getCurrentDraftAtom(atomType, id)
+        postHookAtom <- NotificationLists.createNotificationList(currentDraftAtom)
+        updatedAtom <- publishUpdatedAtom(postHookAtom, req)
+      } yield updatedAtom
+    }
+  }
+
+  def deleteNotificationList(atomType: String, id: String) = AuthAction { req =>
+    APIResponse {
+      for {
+        currentDraftAtom <- getCurrentDraftAtom(atomType, id)
+        postHookAtom <- NotificationLists.deleteNotificationList(currentDraftAtom)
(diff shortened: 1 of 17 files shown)


build.sbt
```diff
@@ -18,7 +18,7 @@ libraryDependencies ++= Seq(
   "com.gu"                   %% "pan-domain-auth-play_2-5"    % "0.4.1",
   "io.circe"                 %% "circe-parser"                % "0.7.0",
   "net.logstash.logback"     %  "logstash-logback-encoder"    % "4.2",
-  "com.chuusai"              %% "shapeless"                   % "2.3.2"
+  "com.gu"                   %% "exact-target-lists"          % "0.1-SNAPSHOT"
```
tomrf1: TODO - update version once released

app/controllers/AtomActions.scala
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  private def getCurrentDraftAtom(atomType: String, id: String): Either[AtomAPIError, Atom] = for {
+    atomType <- validateAtomType(atomType)
+    currentDraftAtom <- atomWorkshopDB.getAtom(previewDataStore, atomType, id)
+  } yield currentDraftAtom
+
+  private def publishUpdatedAtom(atom: Atom, req: UserRequest[_]): Either[AtomAPIError, Atom] = for {
+    updatedAtom <- atomWorkshopDB.publishAtom(publishedDataStore, req.user, updateTopLevelFields(atom, req.user, publish=true))
+    _ <- atomWorkshopDB.updateAtom(previewDataStore, updatedAtom)
```
regiskuckaertz: Should these three actions occur sequentially? I wonder if at least the two below could be triggered at the same time and closed on together; what do you think?

tomrf1: good point, certainly the last 2 can

conf/routes
```diff
@@ -38,3 +38,7 @@ POST    /reindex-preview                         com.gu.atom.play.ReindexControl
 POST    /reindex-publish                         com.gu.atom.play.ReindexController.newPublishedReindexJob()
 GET     /reindex-preview                         com.gu.atom.play.ReindexController.previewReindexJobStatus()
 GET     /reindex-publish                         com.gu.atom.play.ReindexController.publishedReindexJobStatus()
+
+# Custom atom atom actions
```
regiskuckaertz: typo

public/js/components/AtomActions/NotificationList.js
```diff
@@ -0,0 +1,78 @@
...
+  createNotificationList() {
+    this.props.actions.createNotificationList(this.props.atom);
+  }
+
+  deleteNotificationList() {
+    this.props.actions.deleteNotificationList(this.props.atom);
+  }
+
+  render() {
+    const listData = this.getListData(this.props.atom);
```
regiskuckaertz: this belongs in the contructor

tomrf1: it needs to check the status of the list each time the atom data updates - if it's in the constructor this won't happen?

regiskuckaertz: hmm, then it is a piece of state shared between parent-child components. that's a common pitfall in React and there are many ways to solve it, one of them being via properties; perhaps the editorial team prefers one style over another, I'll leave it up to them 👍

PR #157, opened 2017-09-19
Simplified date selection for Timeline atoms

annebyrne:


public/js/components/AtomEdit/CustomEditors/TimelineFields/TimelineItem.js
```diff
@@ -25,14 +25,14 @@ export class TimelineItem extends React.Component {
...
+      date: "",
       body: ""
     };
     return (
       <div className="form__field form__field--nested">
         <ManagedForm data={value} updateData={this.updateItem} onFormErrorsUpdate={this.props.onFormErrorsUpdate} formName="timelineEditor">
-        <ManagedField fieldLocation="date" name="Date" isRequired={true}>
-          <FormFieldDateInput isOutsideRange={() => false}/>
+        <ManagedField fieldLocation="date" name="Date" isRequired={true} customValidation={[isValidTimelineDate]}>
+          <FormFieldTextInput fieldPlaceholder='dd/mm/yyyy'/>
```
tomrf1: Does this end up in the database as milliseconds? I can't work out where the conversion is done

public/js/util/validators.js
```diff
@@ -15,3 +15,13 @@ export const isHttpsUrl = (value) => {
...
   }
 };
+
+export const isValidTimelineDate = (date) => {
+  const stringValue = typeof date === 'string' ? date : '';
+    if (stringValue.match(/^\d{2}[./-]\d{2}[./-]\d{4}$/)) {
+        return Promise.resolve(true);
+    } else {
+      const error = new FieldError('invalid-date', 'Date not in a valid format of dd/mm/yyyy.');
+      return Promise.resolve(error);
```
regiskuckaertz: `Promise.reject(error)`?

public/js/util/validators.js
```diff
@@ -15,3 +15,13 @@ export const isHttpsUrl = (value) => {
     return Promise.resolve(error);
   }
 };
+
+export const isValidTimelineDate = (date) => {
```
regiskuckaertz: terse with 
```
isValidTimelineDate = date => String(date).match(/^\d{2}[./-]\d{2}[./-]\d{4}$/) 
  ? Promise.resolve(true)
  : Promise.reject(...)
```

public/js/util/validators.js
```diff
@@ -15,3 +15,13 @@ export const isHttpsUrl = (value) => {
...
+  const stringValue = typeof date === 'string' ? date : '';
+    if (stringValue.match(/^\d{2}[./-]\d{2}[./-]\d{4}$/)) {
+        return Promise.resolve(true);
+    } else {
+      const error = new FieldError('invalid-date', 'Date not in a valid format of dd/mm/yyyy.');
```
regiskuckaertz: Let's output the value of `date` too, it might help debug problems in the future.

public/js/util/validators.js
annebyrne: I followed the current working example of `isHttpsUrl` for the CTA editor which uses `Promise.resolve(error)`.

public/js/components/AtomEdit/CustomEditors/TimelineFields/TimelineItem.js
```diff
@@ -22,17 +25,24 @@ export class TimelineItem extends React.Component {
     this.props.onUpdateField(item);
   }
 
+  changeDate = (date) => {
+    console.log("updating date woohooo!");
```
regiskuckaertz: 😆

public/js/components/FormFields/FormFieldDateTextInput.js
```diff
@@ -0,0 +1,58 @@
...
+    static propTypes = {
+        fieldLabel: PropTypes.string,
+        fieldName: PropTypes.string,
+        fieldValue: PropTypes.number,
+        fieldErrors: PropTypes.arrayOf(errorPropType),
+        formRowClass: PropTypes.string,
+        onUpdateField: PropTypes.func
+    };
+
+    milisecondsToString = (ms) => {
```
regiskuckaertz: this is a static method, you can move it out of the class and get rid of the annoying `this` juggling.

public/js/components/FormFields/FormFieldDateTextInput.js
```diff
@@ -0,0 +1,58 @@
...
+        moment(ms).format("yyyy-MM-dd");
+    }
+
+    state = {
```
regiskuckaertz: whoa 😮  I have never seen this construct; is it legal? Usually `state` is initialised in the constructor, as the constructor is the starting point for using `props`.

annebyrne: Yup and we had to do it this way otherwise it breaks 😱  this has been an _eye-opening_ PR....

tomrf1: it's certainly used elsewhere in atom-workshop!

regiskuckaertz: so much for React and JavaScript classes 😝  well done 👍

shaundillon: @regiskuckaertz - Only because we're using https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-class-properties

regiskuckaertz:
Great! So is this all that is needed to include that shiny new date picker? Isn't there a third party script to include somewhere?

annebyrne:
Nope this is it! ... though its a deceptively small PR. This has been painful :joy: serious shout out to @shaundillon for tolerating my (many) questions!

tomrf1:
👍 💯

regiskuckaertz:
👏

public/js/components/FormFields/FormFieldDateTextInput.js
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+
+    render() {
+
+        return (
+            <div className={this.props.formRowClass || "form__row"}>
+                {this.props.fieldLabel ? <label htmlFor={this.props.fieldName} className="form__label">{this.props.fieldLabel}</label> : false}
+                <input
+                    type="date"
```
shaundillon: Just for Posterity, this is behind a feature switch in FF<55 but falls back to a text field anyway.

annebyrne: Yes! Rationale for going with this method is a combination of time already spent on this, complexity of task and just getting it ready for launch across newsroom, where majority use Chrome.

tomrf1: Yeah this doesn't work at all in my version of FF, it's impossible to enter text :(

PR #159, opened 2017-09-27
Link answers to questions

tomrf1:
A very basic UI for linking an atom ID or content ID to a question. 
The answers are added at the question-level, rather than the atom-level.

TODO:
1. Perhaps in future it could help the user search for the atom/content so that they don't have to use IDs.
2. Trigger email send after linking an answer

This work took way longer than expected because:
1. I discovered an issue with deserialization to scrooge classes which can sometimes cause data to be lost, hence the fezziwig version bump (https://github.com/guardian/fezziwig/pull/16)
2. I've 'fixed' a long-standing issue with `FormFieldArrayWrapper` which causes updates to be sent back to the server immediately when the user clicks "add". This leads to error messages because `undefined` is appended. The solution is quite hacky, but I seem to have fixed it by using internal state for new items.

With no answers:
picture 17

After clicking add:
picture 18

With an answer:
picture 19

### DIFF
--- a/public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Answer.js
+++ b/public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Answer.js
@@ -1,0 +1,49 @@
+import React, { PropTypes } from 'react';
+import {ManagedForm, ManagedField} from '../../../ManagedEditor';
+import FormFieldTextInput from '../../../FormFields/FormFieldTextInput';
+import SearchSelectBox from '../../../FormFields/SearchFields/SearchSelectBox';
+import _get from 'lodash/fp/get';
+
+export class StoryQuestionsAnswer extends React.Component {
+  static propTypes = {
+    fieldLabel: PropTypes.string,
+    fieldName: PropTypes.string,
+    fieldValue: PropTypes.shape({
+      answerId: PropTypes.string,
+      answerType: PropTypes.string
+    }),
+    fieldPlaceholder: PropTypes.string,
+    onUpdateField: PropTypes.func,
+    onFormErrorsUpdate: PropTypes.func
+  };
+
+  onUpdate = (answer) => {
+    const answerWithType = Object.assign({}, answer, {
+      answerType: answer.answerType || "ATOM"
+    });
+
+    this.props.onUpdateField(answerWithType);
+  }
+
+  getPlaceholder = (answerType) => answerType === "CONTENT"
+    ? "e.g. world/2017/sep/21/an-explainer-article"
+    : "e.g. atom/guide/d12c3782-2d4b-4335-9701-830ac29c7d3b";
+
+  render () {
+    const answerType = _get(this.props, "fieldValue.answerType", "ATOM");
+    const placeholder = this.getPlaceholder(answerType);
+
+    return (
+      <div className="form__field form__field--nested">
+        <ManagedForm data={this.props.fieldValue} updateData={this.onUpdate} onFormErrorsUpdate={this.props.onFormErrorsUpdate} formName="storyquestionsEditor">
+          <ManagedField fieldLocation="answerType" name="Answer type" isRequired={true}>
+            <SearchSelectBox selectValues={["ATOM", "CONTENT"]} isRequired={true}/>
+          </ManagedField>
+          <ManagedField fieldLocation="answerId" name="Answer ID" isRequired={true}>
+            <FormFieldTextInput fieldPlaceholder={placeholder}/>
+          </ManagedField>
+        </ManagedForm>
+      </div>
+    );
+  }
+}
--- a/public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
+++ b/public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
@@ -1,6 +1,8 @@
 import React, { PropTypes } from 'react';
 import {ManagedForm, ManagedField} from '../../../ManagedEditor';
 import FormFieldTextInput from '../../../FormFields/FormFieldTextInput';
+import FormFieldArrayWrapper from '../../../FormFields/FormFieldArrayWrapper';
+import {StoryQuestionsAnswer} from './Answer';
 import uuidv4 from 'uuid/v4';
 
 export class StoryQuestionsQuestion extends React.Component {
@@ -9,7 +11,8 @@ export class StoryQuestionsQuestion extends React.Component {
     fieldName: PropTypes.string,
     fieldValue: PropTypes.shape({
       questionId: PropTypes.string,
-      questionText: PropTypes.string
+      questionText: PropTypes.string,
+      answers: PropTypes.array
     }),
     fieldPlaceholder: PropTypes.string,
     onUpdateField: PropTypes.func,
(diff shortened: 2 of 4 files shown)


regiskuckaertz:
Is the ID alone sufficient or the prefix (`atom/guide`) necessary for this to work?

public/js/components/FormFields/FormFieldArrayWrapper.js
```diff
@@ -18,27 +18,51 @@ export default class FormFieldArrayWrapper extends React.Component {
...
+  //We store new items in state to avoid sending invalid items back to the server before they're ready
+  state = {
+    newItems: []
+  }
+
   onAddClick = () => {
-    const existingValues = this.props.fieldValue || [];
-    this.props.onUpdateField(existingValues.concat([undefined]));
+    this.setState({
+      newItems: this.state.newItems.concat([undefined])
```
regiskuckaertz: Why the `undefined`? 😮

tomrf1:
Good question. Really the ID should be `atom/guide/a8je....`, since that's what you query capi for. @mchv and I have talked in the past about migrating to always using the full ID, though it's a bit painful now

public/js/components/FormFields/FormFieldArrayWrapper.js
```diff
@@ -18,27 +18,51 @@ export default class FormFieldArrayWrapper extends React.Component {
...
 
   renderValue(value, i) {
 
     const updateFn = (newValue) => {
-      //Find the value in the array to change
-      const newFieldValue = this.props.fieldValue.map((oldValue) => {
-        return value === oldValue ? newValue : oldValue;
-      });
-      this.props.onUpdateField(newFieldValue);
+      if (this.state.newItems.indexOf(value) > -1) {
```
regiskuckaertz: `if(this.state.newItems.includes(value))`

public/js/components/FormFields/FormFieldArrayWrapper.js
regiskuckaertz: I see, it's like a placeholder for future values.

mchv:
@tomrf1 Tom is correct, the id of the atom is `atom/guide/a8je....`. I think we still need to migrate from the `UUID` generated in atom workshop to a sensible slug name as we do for articles

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
```diff
@@ -25,12 +28,19 @@ export class StoryQuestionsQuestion extends React.Component {
   }
 
   render () {
+    const answersCount = (this.props.fieldValue && this.props.fieldValue.answers) ? this.props.fieldValue.answers.length : 0;
```
regiskuckaertz: Hmm, so the underlying array wrapper mutates this variable. That is where I find React lacking, there is no combinator to signal state changes upstream 😩

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Answer.js
```diff
@@ -0,0 +1,48 @@
...
+
+    this.props.onUpdateField(answerWithType);
+  }
+
+  getPlaceholder = (answerType) => answerType === "CONTENT"
+    ? "e.g. world/2017/sep/21/an-explainer-article"
+    : "e.g. atom/guide/d12c3782-2d4b-4335-9701-830ac29c7d3b";
+
+  render () {
+    const answerType = this.props.fieldValue && this.props.fieldValue.answerType ? this.props.fieldValue.answerType : "Atom";
```
regiskuckaertz: Not uppercased?

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Answer.js
```diff
@@ -0,0 +1,48 @@
...
+      answerId: PropTypes.string,
+      answerType: PropTypes.string
+    }),
+    fieldPlaceholder: PropTypes.string,
+    onUpdateField: PropTypes.func,
+    onFormErrorsUpdate: PropTypes.func
+  };
+
+  onUpdate = (answer) => {
+    const answerWithType = Object.assign({}, answer, {
```
regiskuckaertz: Just curious: if the field is required, why do you need to override `answerType`?

public/js/components/FormFields/FormFieldArrayWrapper.js
regiskuckaertz: Also you have access to `i` from `renderValue`, so in theory you shouldn't need to do this. You can just then test if `this.state.newItems[i] === undefined`.

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Answer.js
```diff
@@ -0,0 +1,48 @@
...
+      answerType: answer.answerType || "ATOM"
+    });
+
+    this.props.onUpdateField(answerWithType);
+  }
+
+  getPlaceholder = (answerType) => answerType === "CONTENT"
```
annebyrne: What's the logic behind the values for the placeholders here?

tomrf1: they're just examples of a content ID and atom ID

annebyrne:
Nice fix on the `ArrayWrapper`!!

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Answer.js
tomrf1: 💯

public/js/components/FormFields/FormFieldArrayWrapper.js
tomrf1: Yep, and this was the cause of the problems before - it was being immediately propagated back up and sent to the server.
Interestingly, we accidentally worked around this for e.g. timeline because the child Scribe component immediately updates the props and overwrites the previous prop update (the one with `undefined`), before it can be sent to the server

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Answer.js
tomrf1: This is effectively setting a default value if it hasn't been set. There's probably a nicer way to do this...

public/js/components/FormFields/FormFieldArrayWrapper.js
tomrf1: hmm, `i` is the index in the concatenation of `this.props.fieldValue` and `this.state.newItems`.
In fact, I've just tried this with `if (value === undefined)` and it works. This is quite nasty

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
shaundillon: Lodash is included, so you can use [`_.get`](https://lodash.com/docs/4.17.4#get) to make this cleaner.

tomrf1: 💯 amazing

shaundillon: In a way, yes, but realistically, the underlying array wrapper is just passing it's new value to the store, which is then propagated down.

public/js/components/FormFields/FormFieldArrayWrapper.js
```diff
@@ -18,27 +18,51 @@ export default class FormFieldArrayWrapper extends React.Component {
...
+
+        this.props.onUpdateField(newFieldValue);
+      }
     };
 
     const removeFn = (removeIndex) => {
-      const newFieldValue = this.props.fieldValue.filter((value, currentIndex) => {
-        return currentIndex !== removeIndex;
-      });
+      if (this.props.fieldValue && this.props.fieldValue.length > removeIndex) {
```
shaundillon: `_.get('this.props.fieldValue', removeIndex)`

shaundillon: `_.get('this.props', 'fieldValue[removeIndex]')` _should_ work

tomrf1: I'm not sure why, but I can't get this to work :(

regiskuckaertz: `_.get('this.props.fieldValue', 'length', -1) > removedIndex` ?

PR #161, opened 2017-09-28
Have optional fields in Timeline use space rather than hyphens

annebyrne:


public/js/components/AtomEdit/CustomEditors/TimelineFields/TimelineItem.js
```diff
@@ -24,9 +24,9 @@ export class TimelineItem extends React.Component {
 
   render () {
     const value = this.props.fieldValue || {
-      title: "-",
+      title: " ",
```
shaundillon: I'm not sure this is valid in Dynamo, which is why a dash was used.

annebyrne: Gaaaaaaaah 😱 Thanks for spotting. I'll take a look.

annebyrne: So from messing around with the `CODE` & `DEV` DynamoDB tables it seems to accept the spaces. Do you know where this was covered in a PR or anywhere else?

shaundillon: @susiecoleman and @philmcmahon Any idea? ^

tomrf1: Dynamo doesn't allow an empty string...not sure if a space will make it in

annebyrne: It seems to work from modifying this atom directly in Dynamo: https://atomworkshop.code.dev-gutools.co.uk/atoms/timeline/07a75b03-92fd-4c78-9052-ccbb25c478d4/edit

What do you think @shaundillon ?

shaundillon: Ahh, that's what I was thinking. No idea why a dash was chosen over a space then. Maybe for visibility?

tomrf1: 💯 much better than the dash!

annebyrne:
This seems to be working in `CODE` so I am going to go ahead and ship if that's OK? @tomrf1 @shaundillon

shaundillon:
Yep, s'all good with me

tomrf1:
👍

PR #163, opened 2017-09-29
Add description field to add context for Timeline atoms

annebyrne:
I plan on deploying this to `CODE` and checking with Chris M that this solves their use case.

### DIFF
--- a/public/js/components/AtomEdit/CustomEditors/TimelineEditor.js
+++ b/public/js/components/AtomEdit/CustomEditors/TimelineEditor.js
@@ -3,6 +3,7 @@ import FormFieldArrayWrapper from '../../FormFields/FormFieldArrayWrapper';
 import {TimelineItem} from './TimelineFields/TimelineItem';
 import {ManagedField, ManagedForm} from '../../ManagedEditor';
 import {atomPropType} from '../../../constants/atomPropType';
+import FormFieldsScribeEditor from '../../FormFields/FormFieldScribeEditor';
 
 
 export class TimelineEditor extends React.Component {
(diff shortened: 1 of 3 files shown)


public/js/components/FormFields/FormFieldScribeEditor.js
```diff
@@ -27,9 +28,13 @@ export default class FormFieldsScribeEditor extends React.Component {
...
   wordCount = text => text.trim().replace(/<(?:.|\n)*?>/gm, '').split(/\s+/).filter(_ => _.length !== 0).length;
 
+  isTooLong = (wordCount) => this.props.suggestedLength && wordCount > this.props.suggestedLength;
+
   renderWordCount = () => {
-    const wordCount = this.wordCount(this.props.fieldValue);
-    const tooLong = this.props.suggestedLength && wordCount > this.props.suggestedLength;
+    const wordCount = this.props.fieldValue? this.wordCount(this.props.fieldValue) : 0;
+
+    const tooLong = this.isTooLong(wordCount);
```
regiskuckaertz: Is this ever used?

annebyrne: I believe so, two lines down:

`{tooLong ? <span className="form__message__text--error"> (too long)</span>: false}`

regiskuckaertz: oops, I was looking at the diff only, now I can see it in the full source :-)

public/js/components/AtomEdit/CustomEditors/TimelineEditor.js
```diff
@@ -14,12 +15,16 @@ export class TimelineEditor extends React.Component {
   }
 
   render () {
+
     return (
       <div className="form">
         <ManagedForm data={this.props.atom} updateData={this.props.onUpdate} onFormErrorsUpdate={this.props.onFormErrorsUpdate} formName="timelineEditor">
-          <ManagedField fieldLocation="data.timeline.events" name="Events">
+          <ManagedField fieldLocation="description" name="Description" isRequired={false}>
```
tomrf1: How about setting `name="Description (optional)"` instead of adding the new `fieldHelperLabel` prop?

annebyrne:
Tested on `CODE` so going to go ahead and merge

PR #165, opened 2017-10-10
Add commissioning desks to snippets

regiskuckaertz:
Much simpler than I expected thanks to the tag field 😃  It's coming through quite nicely:

screen shot 2017-10-10 at 13 45 45

NB: I've added the ability to filter by type in the tag search, it is not returning any result in preview but I tested with CAPI prod and works fine!

### DIFF
--- a/public/js/components/AtomEdit/CustomEditors/GuideEditor.js
+++ b/public/js/components/AtomEdit/CustomEditors/GuideEditor.js
@@ -31,7 +31,7 @@ export class GuideEditor extends React.Component {
           </ManagedField>
           <ManagedField fieldLocation="commissioningDesks" name="Commissioning desks" updateData={this.onDesk}>
             <FormFieldArrayWrapper>
-              <FormFieldTagPicker />
+              <FormFieldTagPicker tagType="tracking" />
             </FormFieldArrayWrapper>
           </ManagedField>
         </ManagedForm>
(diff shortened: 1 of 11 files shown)


regiskuckaertz:
Mandatory screenshot:

screen shot 2017-10-10 at 14 02 21

public/js/components/AtomEdit/CustomEditors/GuideEditor.js
```diff
@@ -28,6 +29,11 @@ export class GuideEditor extends React.Component {
               <GuideItem onFormErrorsUpdate={this.props.onFormErrorsUpdate} />
             </FormFieldArrayWrapper>
           </ManagedField>
+          <ManagedField fieldLocation="commissioningDesks" name="Commissioning desks" updateData={this.onDesk}>
```
tomrf1: Where is `onDesk` coming from?

regiskuckaertz: ha! good catch, I removed it but forgot to remove the attribute too.

tomrf1:
👍 looks good, though I wonder if this field should be in the same section as the title?

regiskuckaertz:
Oh, I did not think about this one. So it would be common across all atoms, right?

tomrf1:
The field is at the top level of the atom model, though perhaps it will only be used for snippets. Might be worth asking Chris M/W if they have an opinion.

regiskuckaertz:
Mariot spoke and so will I move that as you recommended 🌟

regiskuckaertz:
There it s:

screen shot 2017-10-10 at 16 39 22

tomrf1:
Perfect! 👍 

picture 8

PR #167, opened 2017-10-11
Missed timeline item is not missed anymore

regiskuckaertz:
I added the message to the description field but not the individual items.

🤔  we should also think about the "length" of the snippet as a whole; it is definitely _not_ fine to create a timeline of 20 items.


tomrf1:
👍 
Agreed, something to discuss with Editorial

PR #168, opened 2017-10-12
Remove explainer atom

mchv:
One can not create anymore old explainer atom, see associated change in the [explainer maker tool](https://github.com/guardian/explain-maker/pull/169) so remove all references to it in atom workshop.

PR #169, opened 2017-10-12
filter non recognised atom type

mchv:
Following #168 we should not try to create an atom if not one supported


mchv:
tested locally and on `PROD`

tomrf1:
👍

PR #170, opened 2017-10-13
Integrate workflow

Reettaphant:
- Integrate workflow to atom workshop - similar to the support provided by media atom maker
- Allow for tracking supported atoms. 
- If atom type is unsupported, don't provide any workflow support.

screen shot 2017-10-13 at 11 08 25
 from atom workshop and display their status if tracked in workflow

### DIFF
--- a/public/js/util/getProductionOffice.js
+++ b/public/js/util/getProductionOffice.js
@@ -1,0 +1,34 @@
+import moment from 'moment';
+
+function isBetween({value, lower, upper}) {
+  return value >= lower && value <= upper;
+}
+
+function getTimezoneOffset() {
+  const offset = moment().utcOffset();
+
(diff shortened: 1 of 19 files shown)


public/js/util/getProductionOffice.js
```diff
@@ -0,0 +1,34 @@
+import moment from 'moment';
+
+function isBetween({value, lower, upper}) {
```
shaundillon: The destructuring here is a bit redundant, as there's nothing stopping the calls below just sending three arguments

Reettaphant: I like it because it it's an easy way to make explicit the argument names in the calls to `isBetween` (but as nice as it is, I can't claim credit for it - I copied this from mam).

PR #172, opened 2017-11-01
Add ability to delete atoms

shaundillon:
Only available to users with the `delete_atom` permission.

Adds Server-side and client side permissions checks to AW


shaundillon:
[Delete atom button](https://trello.com/c/rw307kwM/210-delete-atom-button)

PR #173, opened 2017-11-01
Replaced all type icons

akemitakagi:
Changed all the type icons and changed padding around the icons.


tomrf1:
👍 tested in CODE

PR #176, opened 2017-11-10
Suggestions page

tomrf1:
1. gets most-read 20 articles from ophan
2. uses the tags to get suggested snippet atoms from Targeting api
3. for each article which does not already have a snippet embedded, displays a list of suggested atoms

picture 16

### DIFF
--- a/public/js/actions/AtomActions/getSuggestionsForLatestContent.js
+++ b/public/js/actions/AtomActions/getSuggestionsForLatestContent.js
@@ -1,8 +1,9 @@
 import {getLatestContent} from '../../services/capi';
-import {fetchTargetsForTags} from '../../services/TargetingApi';
+import {fetchTargetsForTag} from '../../services/TargetingApi';
 import AtomsApi from '../../services/AtomsApi';
+import {atomPropType} from '../../constants/atomPropType.js';
 import {logError} from '../../util/logger';
-
+import {PropTypes} from 'react';
 
 function requestSuggestionsForLatestContent() {
   return {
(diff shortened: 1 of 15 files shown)


public/js/actions/AtomActions/getSuggestionsForLatestContent.js
```diff
@@ -0,0 +1,148 @@
...
+        const urls = filtered.map(target => target.url);
+        if (filtered.length > 0) {
+          tagToTargetAtoms[tag] = (tagToTargetAtoms[tag]) ? tagToTargetAtoms[tag].tag(urls) : urls;
+        }
+        return Promise.resolve();
+      });
+  };
+
+  //Sequentially fetch targets for each tag
+  return tags.reduce((p, tag) => p.then(() => fetch(tag)), Promise.resolve())
```
regiskuckaertz: You can thread `tapToTargetAtoms` in the promise chain if you prefer that style:
```
const fetch = (tag, tapToTargetAtoms) => fetchTargetsForTag(tag)
  .then(targets =>
    ({ ...tapToTargetAtoms,
      [tag]: target.filter(...).map(...)
    }));

return tags.reduce((p, tag) => p.then(ttta => fetch(tag, ttta)), Promise.resolve({});
```

public/js/actions/AtomActions/getSuggestionsForLatestContent.js
```diff
@@ -0,0 +1,148 @@
...
+    .then(() => {
+      return Promise.resolve(tagToTargetAtoms);
+    });
+}
+
+//Deduplicate the targeting urls and retrieve the atoms
+function getAtomUrlToAtom(tagToUrls) {
+  const atomUrlToAtom = {};
+
+  const atomUrls = distinct(flatten(Object.keys(tagToUrls).map(tag => tagToUrls[tag])));
```
regiskuckaertz: `const atomUrls = distinct(flatten(Object.values(tagToUrls)))` should work

public/js/actions/AtomActions/getSuggestionsForLatestContent.js
```diff
@@ -0,0 +1,148 @@
...
+  //Sequentially fetch targets for each tag
+  return tags.reduce((p, tag) => p.then(() => fetch(tag)), Promise.resolve())
+    .then(() => {
+      return Promise.resolve(tagToTargetAtoms);
+    });
+}
+
+//Deduplicate the targeting urls and retrieve the atoms
+function getAtomUrlToAtom(tagToUrls) {
+  const atomUrlToAtom = {};
```
regiskuckaertz: You can move that var into the last `.then` callback

public/js/services/TargetingApi.js
```diff
@@ -62,3 +62,37 @@ export const deleteTarget = (targetId) => {
...
   );
 };
+
+export const fetchTargetsForTags = (tags) => {
+
+  const store = getStore();
+  const state = store.getState();
+  const targetingUrl = state.config.targetingUrl;
+
+  const tagsParams = tags.map(tag => `tags=${tag}`).join('&');
```
regiskuckaertz: `tags=${encodeURIComponent(tag)}`?

public/js/actions/AtomActions/getSuggestionsForLatestContent.js
```diff
@@ -0,0 +1,148 @@
...
+function getTagToTargetUrls(tags) {
+  const tagToTargetAtoms = {};
+
+  const fetch = (tag) => {
+    return fetchTargetsForTag(tag)
+      .then(targets => {
+        const filtered = targets.filter(target => target.url.includes("/atom/"));
+        const urls = filtered.map(target => target.url);
+        if (filtered.length > 0) {
+          tagToTargetAtoms[tag] = (tagToTargetAtoms[tag]) ? tagToTargetAtoms[tag].tag(urls) : urls;
```
regiskuckaertz: Regardless of which style you prefer, I don't think the consequent is every going to be called here. As each tag is unique, you can just write `tagToTargetAtoms[tag] = urls`.

regiskuckaertz:
Solid stuff! It all looks good to me, except for that last comment with urls 👍

public/js/actions/AtomActions/getSuggestionsForLatestContent.js
tomrf1: good point, I'm really not sure why I wrote that...

PR #177, opened 2017-11-13
Speed up suggestions page

tomrf1:
This page will only get suggestions for <= 20 articles, so it's generally quicker to send 1 query per article to the targeting api (rather than 1 per tag as before).

### DIFF
--- a/public/js/actions/AtomActions/getSuggestionsForLatestContent.js
+++ b/public/js/actions/AtomActions/getSuggestionsForLatestContent.js
@@ -34,13 +34,19 @@ function errorReceivingSuggestionsForLatestContent(error) {
 const distinct = array => array.filter((value, index, self) => self.indexOf(value) === index);
 const flatten = array => [].concat.apply([], array);
 
-//Returns an array of target atom urls for the given tags
-function getTargetUrls(tags) {
+//Returns set of all tags in the given content
+function getTags(contentArray) {
+  return distinct(flatten(
+    contentArray.map(content => content.tags.filter(tag => tag.type === "keyword"))
+  ));
+}
+
+//Returns an array of target atoms for the given tags
+function getTargets(tags) {
   const tagIds = tags.filter(tag => tag.type === "keyword").map(tag => tag.id);
 
   return fetchTargetsForTags(tagIds).then(targets => {
-    const filtered = targets.filter(target => target.url.includes("/atom/"));
-    return filtered.map(target => target.url);
+    return targets.filter(target => target.url.includes("/atom/"));
   });
 }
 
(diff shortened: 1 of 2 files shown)


public/js/actions/AtomActions/getSuggestionsForLatestContent.js
```diff
@@ -34,44 +34,32 @@ function errorReceivingSuggestionsForLatestContent(error) {
...
+    const filtered = targets.filter(target => target.url.includes("/atom/"));
+    return filtered.map(target => target.url);
+  });
+}
 
-  //Sequentially fetch targets for each tag
-  return tags.reduce((p, tag) => p.then(() => fetch(tag)), Promise.resolve())
-    .then(() => Promise.resolve(tagToTargetAtoms));
+const AllSnippets = ["guide","qanda","timeline","profile"];
+function isSnippet(atomType) {
```
regiskuckaertz: This kind of code is all over the place. I wonder if adding an `isSnippet` flag to all atoms would make our lives easier.

tomrf1: you mean in the model? It could be useful, though we'd have to apply it to all existing snippets

regiskuckaertz:
Can this be improved even further by folding/unfolding tags so that you only run a single query?

public/js/actions/AtomActions/getSuggestionsForLatestContent.js
regiskuckaertz: Yeah it's just a thought, but since snippet seem to be part of our everyday vocabulary, and we might grow the family in the future, why not? I wonder what @mchv thinks about that. That flag would need to be added to `Atom` in order to be really effective.

mchv: I would not add that to the model but in the client if this is useful with an extension method on the `Atom` class

tomrf1:
@regiskuckaertz I was about to say no, but you're right it is possible to do in 1 query because the targeting response does include the tags for each target!

public/js/actions/AtomActions/getSuggestionsForLatestContent.js
mchv: i.e 
```scala
implicit class RichAtom(val atom: Atom) extends AnyVal {
   def isSnippet =atom.type == AtomType.Qanda || atom.type == AtomType.Profile || atom.type == AtomType.Guideline ||atom.type == AtomType.Timeline
```

regiskuckaertz: Yes, that makes perfect sense.

tomrf1:
now updated to make only 1 request to targeting api

regiskuckaertz:
👍

PR #178, opened 2017-11-15
Flexible date formats for Timelines

annebyrne:
Allow users to select relevant date format for events and date ranges if required.

Tested on `CODE`

screen shot 2017-11-15 at 11 42 59

### DIFF
--- a/public/js/components/AtomEdit/CustomEditors/TimelineFields/TimelineItem.js
+++ b/public/js/components/AtomEdit/CustomEditors/TimelineFields/TimelineItem.js
@@ -3,30 +3,74 @@ import {ManagedForm, ManagedField} from '../../../ManagedEditor';
 import FormFieldTextInput from '../../../FormFields/FormFieldTextInput';
 import FormFieldsScribeEditor from '../../../FormFields/FormFieldScribeEditor';
 import FormFieldDateTextInput from '../../../FormFields/FormFieldDateTextInput';
+import FormFieldRadioButtons from '../../../FormFields/FormFieldRadioButtons';
+import FormFieldCheckbox from "../../../FormFields/FormFieldCheckbox";
 
 export class TimelineItem extends React.Component {
   static propTypes = {
     fieldLabel: PropTypes.string,
     fieldName: PropTypes.string,
     fieldValue: PropTypes.shape({
-      title: PropTypes.string,
-      date: PropTypes.number,
-      body: PropTypes.string
+      title: PropTypes.string.isRequired,
+      date: PropTypes.number.isRequired,
+      toDate: PropTypes.number,
+      body: PropTypes.string,
+      dateFormat: PropTypes.string
     }),
     fieldPlaceholder: PropTypes.string,
     onUpdateField: PropTypes.func,
     onFormErrorsUpdate: PropTypes.func
   };
 
+    state = {
+        dateRangeRequired: typeof this.props.fieldValue === PropTypes.shape({
+            title: PropTypes.string,
+            date: PropTypes.number,
+            toDate: PropTypes.number,
+            body: PropTypes.string,
+            dateFormat: PropTypes.string
+        }),
+    };
+
+
+  dateFormats = [
+      {
+          value: 'day-month-year',
+          name: 'Calendar date e.g. Monday 13th July 2017'
+      },
+      {
+          value: 'month-year',
+          name: 'Month & year e.g. July 2017'
+      },
+      {
+          value: 'year',
+          name: 'Year e.g. 2017'
+      }
+  ]
+
+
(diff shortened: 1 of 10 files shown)


public/js/components/AtomEdit/CustomEditors/TimelineFields/TimelineItem.js
```diff
@@ -3,30 +3,74 @@ import {ManagedForm, ManagedField} from '../../../ManagedEditor';
...
+      body: PropTypes.string,
+      dateFormat: PropTypes.string
     }),
     fieldPlaceholder: PropTypes.string,
     onUpdateField: PropTypes.func,
     onFormErrorsUpdate: PropTypes.func
   };
 
+    state = {
+        dateRangeRequired: typeof this.props.fieldValue === PropTypes.shape({
```
tomrf1: what's happening here, is it checking for the existence of `toDate`?

annebyrne: Yes it is. On first render, `fieldValue` is undefined so this is why `this.props.fieldValue.toDate` doesn't work and we require a more verbose solution

annebyrne:
There is a bug in React which means that multiple radio button lists will cancel the `checked` attribute of all buttons save one. Because of this, we've chosen to go with a select box for now until the problem is resolved.

It should hopefully be sorted soon so the relevant code will be uncommented and removed.

public/js/components/AtomEdit/CustomEditors/TimelineFields/TimelineItem.js
regiskuckaertz: have you tried `this.props.fieldValue && 'toDate' in this.props.fieldValue`?

public/js/components/AtomEdit/CustomEditors/TimelineFields/TimelineItem.js
```diff
@@ -3,30 +3,85 @@ import {ManagedForm, ManagedField} from '../../../ManagedEditor';
...
+       //  Remove ID in dateFormats array as will be superfluous
+       //
+       //  See this PR for further info: https://github.com/facebook/react/pull/11227
+       //
+       //  The radio button code has been left in for now as React devs are working on this bug.
+       //  React cannot handle multiple radio button lists and leaves only one button visibly checked
+       //  regardless of the props. This is only a bug of UX, and the data is being stored correctly
+       //  on the atom. We've chosen to release this feature with a select so to speed up delivery.
+
+dateFormats = [
```
regiskuckaertz: Best put these out of the component or at least as a static property.

public/js/components/AtomEdit/CustomEditors/TimelineFields/TimelineItem.js
annebyrne: It works but produces a few warnings. I think I'm going to stick with the above for now.

public/js/components/FormFields/FormFieldRadioButtons.js
```diff
@@ -1,40 +1,51 @@
...
   }
 
-  renderButton(value, i) {
+  renderButton(value, label, i) {
     return (
-      <div key={i} className="form__group--radio">
-        <input className="form__radio-btn" type="radio" name={this.props.fieldName} value={value} checked={this.props.fieldValue === value ? 'checked' : false} onChange={this.onUpdate} />
-        <span className="form__label form__label--radio">{value}</span>
+      <div key={i+uuidv4()} className="form__group--radio">
+        <input key={uuidv4()} className="form__radio-btn" type="radio" value={value} checked={this.props.fieldValue === value ? 'checked' : false} onChange={this.onUpdate} name={this.props.fieldName} />
```
tomrf1: we can get rid of these uuids now that we've finished debugging

public/js/components/AtomEdit/CustomEditors/TimelineFields/TimelineItem.js
```diff
@@ -35,6 +90,25 @@ export class TimelineItem extends React.Component {
           <ManagedField fieldLocation="date" name="Date" isRequired={true}>
             <FormFieldDateTextInput/>
           </ManagedField>
+          <FormFieldCheckbox fieldName="ranged-date" fieldLabel="Want a date range?" fieldValue={this.state.dateRangeRequired} onUpdateField={this.updateDateRange.bind(this)}/>
+          <ManagedField isRequired={false} fieldLocation="toDate">
```
regiskuckaertz: Another option would be to render this field only if `state.dateRangeRequired`

regiskuckaertz:
Nice!

tomrf1:
👍 🎉

PR #182, opened 2017-11-27
Set word limit on Snippets

annebyrne:
This sets a word limit on Q&As of 250 words, and on Guides, Timelines & Profiles of 400 words across multiple blocks within the atom.

### DIFF
--- a/public/js/components/AtomEdit/CustomEditors/QAndAFields/QAItem.js
+++ b/public/js/components/AtomEdit/CustomEditors/QAndAFields/QAItem.js
@@ -2,6 +2,8 @@ import React, { PropTypes } from 'react';
 import {ManagedForm, ManagedField} from '../../../ManagedEditor';
 import FormFieldsScribeEditor from '../../../FormFields/FormFieldScribeEditor';
 
+const WordLimit = 250;
+
 export class QAItem extends React.Component {
   static propTypes = {
     fieldLabel: PropTypes.string,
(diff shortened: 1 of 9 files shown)


annebyrne:
Note: This seems to still allow atoms over the word count (ie. that have failed the custom validation) to be saved & published as seen in `CODE` DB.

public/js/components/AtomEdit/CustomEditors/QAndAFields/QAItem.js
```diff
@@ -29,7 +29,7 @@ export class QAItem extends React.Component {
       <div className="form__field">
         <ManagedForm data={value} updateData={this.updateItem} onFormErrorsUpdate={this.props.onFormErrorsUpdate} formName="qaEditor">
           <ManagedField fieldLocation="body" name="Answer" isRequired={true}>
-            <FormFieldsScribeEditor showWordCount={true} suggestedLength={150} showToolbar={false} tooLongMsg={"Remember that snippets should be concise"}/>
+            <FormFieldsScribeEditor showWordCount={true} suggestedLength={250} showToolbar={false} tooLongMsg={"You've exceed the word limit of 250 for this atom."}/>
```
regiskuckaertz: `exceeded`?

public/js/util/validators.js
```diff
@@ -14,4 +14,30 @@ export const isHttpsUrl = (value) => {
...
+        } else {
+            const error = new FieldError('too-long', 'You\'ve reached the word limit on this atom type. Please edit your items to less than a total of 400 words.');
+            return Promise.resolve(error);
+        }
+
+    }
+};
+
+function stripHtml(rawBody) {
+  return rawBody.replace(/<{1}[^<>]{1,}>{1}/g,"");
```
regiskuckaertz: 😮

annebyrne:
So the still saving & publishing behaviour seems to also be true for current validation in CTA atoms. @shaundillon is this intended behaviour?

public/js/util/validators.js
```diff
@@ -14,4 +14,30 @@ export const isHttpsUrl = (value) => {
...
     return Promise.resolve(error);
   }
-};
\ No newline at end of file
+};
+
+export const checkItemsUnderWordCount = (values) => {
+    if (values.length > 0) {
+        var wordCount = values.map(function(itemData) {
+            var strippedText = stripHtml(itemData.body);
```
regiskuckaertz: Do you know where `itemData.body` is coming from? I'm thinking, if you can use the browser to do the stripping for you, that's even better (via `element.textContent`)

regiskuckaertz:
LGTM 👍

public/js/util/validators.js
tomrf1: alternative - https://www.npmjs.com/package/striptags

public/js/util/validators.js
tomrf1: This is the string value in the model, so unless we can first turn it into a DOM element I don't think so

PR #183, opened 2017-12-04
Fix timeline range

tomrf1:
Fixes a bug where the `toDate` box is not displayed on page load (the data was still going into the db).

The first commit is just indentation fixes

TODO - there's also a bug that prevents you from removing the toDate once it's been set, but I'll look into that later

### DIFF
--- a/public/js/components/AtomEdit/CustomEditors/TimelineFields/TimelineItem.js
+++ b/public/js/components/AtomEdit/CustomEditors/TimelineFields/TimelineItem.js
@@ -23,45 +23,45 @@ export class TimelineItem extends React.Component {
     onFormErrorsUpdate: PropTypes.func
   };
 
-    state = {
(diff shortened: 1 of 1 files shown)


public/js/components/AtomEdit/CustomEditors/TimelineFields/TimelineItem.js
```diff
@@ -71,14 +65,14 @@ dateFormats = [
 
   updateDateRange = (dateRangeRequired) => {
     this.setState({
-          dateRangeRequired: dateRangeRequired
-      });
-    };
+      dateRangeRequired: dateRangeRequired
+    });
```
regiskuckaertz: Yep, here the content of the `toDate` field should probably be emptied when `!dateRangeRequired`. For another PR.

PR #184, opened 2017-12-04
Allow removal of toDate on timelines

tomrf1:
Currently you can't remove a date range once you've set it.

There must be a nicer way of doing this...

### DIFF
--- a/public/js/components/AtomEdit/CustomEditors/TimelineFields/TimelineItem.js
+++ b/public/js/components/AtomEdit/CustomEditors/TimelineFields/TimelineItem.js
@@ -67,6 +67,16 @@ export class TimelineItem extends React.Component {
     this.setState({
       dateRangeRequired: dateRangeRequired
     });
+
+    //Remove the toDate if no longer set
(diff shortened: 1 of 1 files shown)


public/js/components/AtomEdit/CustomEditors/TimelineFields/TimelineItem.js
```diff
@@ -67,6 +67,16 @@ export class TimelineItem extends React.Component {
     this.setState({
       dateRangeRequired: dateRangeRequired
     });
+
+    //Remove the toDate if no longer set
+    const updated = Object.assign({}, this.props.fieldValue, {
+      title: this.props.fieldValue.title,
+      date: this.props.fieldValue.date,
+      toDate: dateRangeRequired ? this.props.fieldValue.toDate : null,
```
regiskuckaertz: you can leave out everything but `toDate` in this object.

tomrf1: ah, thanks!

PR #185, opened 2017-12-05
use signature version 4 to authenticate requests

Reettaphant:
So that we can fetch configs from encrypted buckets.

### DIFF
--- a/fetch-config.sh
+++ b/fetch-config.sh
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
 
+region="eu-west-1"
+
 mkdir -p ~/.configuration-magic/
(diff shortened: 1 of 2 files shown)


fetch-config.sh
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
 
+region="eu-west-1"
+
 mkdir -p ~/.configuration-magic/
-aws s3 cp s3://guconf-flexible/atom-workshop/atom-workshop.conf ~/.configuration-magic/atom-workshop.conf --profile composer
+aws s3 cp s3://guconf-flexible/atom-workshop/atom-workshop.conf ~/.configuration-magic/atom-workshop.conf --profile composer --region $region
```
akash1810: strange that `s3 cp` needs a region

Reettaphant: Yes, it started demanding that after we encrypted the buckets.

PR #186, opened 2017-12-07
Use atom renderer for default html for reader questions and snippets

regiskuckaertz:
That was added back when we needed a preview of the atom in composer, but since we are using the iframe instead (soon the library), we don't need this anymore.

That will also fix a bug in dotcom/apps where closed reader questions still show up, because `defaultHtml` is spliced in the body of articles.


mchv:
👍

PR #188, opened 2017-12-08
Revert rendering library

regiskuckaertz:
Error seems to be coming from incompatibility between the Twirl version used in the library (1.3.4) and the one used in atom workshop (1.1.1).

Solution: upgrade atom workshop to Play 2.6 first.


tomrf1:
👍

PR #194, opened 2018-01-03
Allow array elements to be reordered

mchv:
I have been stuck on this for more than one day because of the Scribe editor. Basically the rendering the editor component is currently done in a way that does not seems compatible with the state being updated outside of the html input:

-   without `componentDidUpdate` method, even if the state is correctly modified the rendered text will still be the old one. 

-  with `componentDidUpdate` method, when reordering (`move up`, `move down`) the html is correctly updated however, the scribe editor does not work anymore.

I have been able to debug that when an edition is done on the scribe editor, the state is not updated at the time the rendering is done, i.e  `this.props.value` in  `componentDidUpdate` will contain the old state :/ 

@shaundillon

### DIFF
--- a/public/js/components/AtomEdit/AtomEdit.js
+++ b/public/js/components/AtomEdit/AtomEdit.js
@@ -7,11 +7,10 @@ import {GuideEditor} from './CustomEditors/GuideEditor';
 import {ProfileEditor} from './CustomEditors/ProfileEditor';
 import {TimelineEditor} from './CustomEditors/TimelineEditor';
 import EmbeddedAtomPick from './EmbeddedAtomPick';
-
 import {subscribeToPresence, enterPresence} from '../../services/presence';
-
 import AtomEditHeader from './AtomEditHeader';
 import {atomPropType} from '../../constants/atomPropType';
+import {logError} from '../../util/logger';
 
 class AtomEdit extends React.Component {
 
@@ -43,7 +42,7 @@ class AtomEdit extends React.Component {
     try {
         enterPresence(this.props.routeParams.atomType, this.props.routeParams.id);
     } catch (e) {
-      console.warn(e);
+        logError(e);
     } finally {
         this.props.atomActions.updateAtom(newAtom);
     }
--- a/public/js/components/AtomEdit/AtomEdit.js
+++ b/public/js/components/AtomEdit/AtomEdit.js
@@ -40,8 +40,13 @@ class AtomEdit extends React.Component {
   }
 
   updateAtom = (newAtom) => {
-    this.props.atomActions.updateAtom(newAtom);
-    enterPresence(this.props.routeParams.atomType, this.props.routeParams.id);
+    try {
+        enterPresence(this.props.routeParams.atomType, this.props.routeParams.id);
+    } catch (e) {
+      console.warn(e);
+    } finally {
+        this.props.atomActions.updateAtom(newAtom);
+    }
   }
 
   updateFormErrors = (errors) => {
--- a/public/js/components/FormFields/FormFieldArrayWrapper.js
+++ b/public/js/components/FormFields/FormFieldArrayWrapper.js
@@ -65,6 +65,19 @@ export default class FormFieldArrayWrapper extends React.Component {
       }
     };
 
+    const moveInArrayFn = (arr, fromIndex, toIndex) => {
+        arr.splice(toIndex, 0, arr.splice(fromIndex, 1)[0]);
+    };
+    
+    const moveFn = (currentIndex, newIndex) => {
+     /* we only allow to move if item has been updated and indices are within array bounds */
+     if (value !== undefined && this.props.fieldValue && this.props.fieldValue.length > currentIndex && currentIndex >= 0 && this.props.fieldValue.length > newIndex && newIndex >= 0) {
+        const newFieldValue = this.props.fieldValue.slice();
+        moveInArrayFn(newFieldValue, currentIndex, newIndex);
+        this.props.onUpdateField(newFieldValue);
+     }
+    };
+
     const hydratedChildren = React.Children.map(this.props.children, (child) => {
       return React.cloneElement(child, {
         key: `${this.props.fieldName}-${i}`,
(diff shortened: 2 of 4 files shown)


public/js/components/FormFields/FormFieldArrayWrapper.js
```diff
@@ -65,6 +65,19 @@ export default class FormFieldArrayWrapper extends React.Component {
       }
     };
 
+    const moveInArrayFn = (arr, fromIndex, toIndex) => {
+        arr.splice(toIndex, 0, arr.splice(fromIndex, 1)[0]);
```
regiskuckaertz: 😱

That would break when `toIndex > fromIndex` because of the inner mutation. It might be easier and safer to use a separate index for array sorting and implement it using a linked list.

```js
const cons = (h, t) => ({ head: h, tail: t });
const car = l => l.head;
const cdr = l => l.tail;
const nth = (i, l) => l == null ? null : i == 0 ? car(l) : nth(i - 1, cdr(l));
const take = (i, l) => l == null || i == 0 ? null : cons(car(l), take(i - 1, cdr(l)));
const drop = (i, l) => l == null ? null : i == 0 ? l : drop(i - 1, cdr(l));
const append = (l, m) => l == null ? m : cons(car(l), append(cdr(l), m));
const remove = (i, l) => l == null ? null : i == 0 ? cdr(l) : cons(car(l), remove(i - 1, cdr(l)));
const insert = (a, i, l) => l == null || i == 0 ? cons(a, l) : cons(car(l), insert(a, i - 1, cdr(l)));

const moveFn = (l, currentIndex, newIndex) => 
  currentIndex == newIndex ? l : insert(nth(currentIndex, l), Math.max(newIndex - 1, 0), remove(currentIndex, l));
```

and then use that list for displaying the children.

regiskuckaertz:
Worth moving to Prose Mirror?

public/js/components/FormFields/FormFieldArrayWrapper.js
mchv: :confused: this code works absolutely fine  with `toIndex > fromIndex` I have tested it. you can find the same approach on this [stack overflow response](https://stackoverflow.com/a/5306832). Here I did not test the bounds because it is only intended to be use in a function that control the bounds.

I find your example add a lot of complexity and potential for bugs for a very basic feature.

regiskuckaertz: I had to manually write down an example to figure out where I was wrong 😅 It would be clearer if spread into several expressions, in my opinion.

annebyrne:
Behaviour now works as expected, for both scribe and form field elements: 

giphy

public/js/components/FormFields/FormFieldScribeEditor.js
```diff
@@ -68,22 +69,29 @@ export class ScribeEditor extends React.Component {
...
+  };
 
-  componentDidMount() {
-    this.scribe = new Scribe(this.refs.editor);
+  state = {
+    scribeElement: null
+  };
 
-    this.configureScribe();
+  uid = uuidv4();
```
mchv: should be `uuid`

public/js/components/FormFields/FormFieldArrayWrapper.js
```diff
@@ -65,6 +66,30 @@ export default class FormFieldArrayWrapper extends React.Component {
...
+          if (value !== undefined && this.props.fieldValue && this.props.fieldValue.length > currentIndex && currentIndex >= 0 && this.props.fieldValue.length > newIndex && newIndex >= 0) {
+              const arr = this.props.fieldValue.slice();
+              moveInArrayFn(arr, currentIndex, newIndex);
+              this.props.onUpdateField(arr);
+          }
+
+          this.setState({
+            childrenVisible: true
+          });
+
```
mchv: Would be good to add a comment explaining why we need to do this.

mchv:
🎖 Great work @annebyrne !

annebyrne:
Thanks @mchv ! As ever srs props _(heh)_ to @shaundillon for his help during the more head-squeezy moments of this.

public/js/components/AtomEdit/AtomEdit.js
```diff
@@ -40,8 +38,13 @@ class AtomEdit extends React.Component {
   }
 
   updateAtom = (newAtom) => {
-    this.props.atomActions.updateAtom(newAtom);
-    enterPresence(this.props.routeParams.atomType, this.props.routeParams.id);
+    try {
+        enterPresence(this.props.routeParams.atomType, this.props.routeParams.id);
+    } catch (e) {
+      console.warn(e); // eslint-disable-line no-console
```
shaundillon: To get around this there's a `logger` module included. You can see examples in the action creators. e.g. https://github.com/guardian/atom-workshop/blob/master/public/js/actions/AtomActions/createAtom.js#L4

public/js/components/FormFields/FormFieldArrayWrapper.js
```diff
@@ -76,10 +101,24 @@ export default class FormFieldArrayWrapper extends React.Component {
       });
     });
 
+    const returnMoveBtns = () => {
```
shaundillon: `renderMoveButtons` would be more consistent with other examples of these functions in the app

public/js/components/FormFields/FormFieldScribeEditor.js
```diff
@@ -109,14 +117,16 @@ export class ScribeEditor extends React.Component {
       }
     }));
 
+    this.scribe.on('content-changed', this.onContentChange);
+    this.state.scribeElement.innerHTML = this.props.value; // eslint-disable-line react/no-direct-mutation-state
```
shaundillon: You can avoid switching off eslint here by doing something along the lines of: 
```js
// Creates a copy of the scribeElement with the new innerHtml
const newElement = Object.assign({}, this.state.scribeElement, {
  innerHtml: this.props.value
}

// Changes the scribeElement in state
this.setState({
  scribeElement: newElement
});
```

It's less destructive and more reacty too

tomrf1:
💯 great feature

PR #195, opened 2018-01-10
Capi iam preview

tomrf1:
We are in the process of changing how we authorise access to CAPI preview. In the past we have used basic auth (in Concierge), with the usernames/passwords maintained by the CAPI team.
Instead we will use api-gateway to move authorisation out of concierge and avoid having to maintain usernames/passwords.

The api-gateway uses IAM authorisation. This means the app assumes a cross-account role, and signs each (preview) request using this role.

Note - this means you will now need capi credentials to run the app locally

### DIFF
--- a/app/controllers/Support.scala
+++ b/app/controllers/Support.scala
@@ -1,23 +1,28 @@
 package controllers
 
+import com.gu.contentapi.client.IAMSigner
 import config.Config
-import play.api.libs.ws.{WSAuthScheme, WSClient}
+import play.api.libs.ws.WSClient
 import play.api.mvc.Controller
 import play.api.libs.concurrent.Execution.Implicits._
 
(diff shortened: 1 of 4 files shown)


app/controllers/Support.scala
```diff
@@ -1,23 +1,28 @@
...
-    val capiUrl = Config.capiPreviewUrl
+  def previewCapiProxy(path: String) = APIAuthAction.async { request =>
+    val capiUrl = Config.capiPreviewIAMUrl
 
     val url = s"$capiUrl/$path?${request.rawQueryString}"
 
     val req = wsClient
       .url(url)
-      .withAuth(capiPreviewUser, capiPreviewPassword, WSAuthScheme.BASIC)
+      .withHeaders(getHeaders(url): _*)
```
annebyrne: 👍

PR #197, opened 2018-01-17
Fix preview

tomrf1:
This was failing after 15mins because it needs to refresh the credentials!


regiskuckaertz:
So your little utility function will renew it automatically?

tomrf1:
Yep, our library calls `getCredentials` on the provider, which will refresh credentials if expired

PR #198, opened 2018-01-23
Explainers: Add back explainer code

ajwl:
This change allows for legacy explainer atoms to be edited. 
We don't want new explainer atoms to be created, so explainers don't come up as a choice in the Create New Atom section. 
However you can search for them in the main list and open already created ones in an explainer editor. 

Check on CODE: https://atomworkshop.code.dev-gutools.co.uk/

**Doesn't show up in Create New Atom options** 
screen shot 2018-01-23 at 16 38 36

**Does show up for editing and searching** 
screen shot 2018-01-23 at 16 38 51

### DIFF
--- a/public/js/components/AtomEdit/CustomEditors/ExplainerEditor.js
+++ b/public/js/components/AtomEdit/CustomEditors/ExplainerEditor.js
@@ -1,0 +1,25 @@
+import React, {PropTypes} from 'react';
+import {ManagedForm, ManagedField} from '../../ManagedEditor';
+import FormFieldTextInput from '../../FormFields/FormFieldTextInput';
+import FormFieldsScribeEditor from '../../FormFields/FormFieldScribeEditor';
+import {atomPropType} from '../../../constants/atomPropType';
+
+export class ExplainerEditor extends React.Component {
+
+  static propTypes = {
+    atom: atomPropType.isRequired,
+    onUpdate: PropTypes.func.isRequired,
+    onFormErrorsUpdate: PropTypes.func
+  }
+
+  render() {
+    return (<ManagedForm data={this.props.atom} updateData={this.props.onUpdate} onFormErrorsUpdate={this.props.onFormErrorsUpdate} formName="explainerEditor">
+      <ManagedField fieldLocation="data.explainer.title" name="Explainer Title" isRequired={true}>
+        <FormFieldTextInput/>
+      </ManagedField>
+      <ManagedField fieldLocation="data.explainer.body" name="Body">
+        <FormFieldsScribeEditor showWordCount={true} suggestedLength={100}/>
+      </ManagedField>
+    </ManagedForm>);
+  }
+}
--- a/public/js/components/Header/EditHeader.js
+++ b/public/js/components/Header/EditHeader.js
@@ -12,7 +12,7 @@ class EditHeader extends React.Component {
 
   static propTypes = {
     atom: atomPropType,
-    presence: PropTypes.object,
+    presence: PropTypes.bool,
     saveState: PropTypes.object,
     atomActions: PropTypes.shape({
       publishAtom: PropTypes.func.isRequired,
@@ -80,7 +80,7 @@ class EditHeader extends React.Component {
     return (
         <div className="toolbar__container">
           {this.props.presence ? <PresenceIndicator presence={this.props.presence} /> : false}
-            <button disabled={atomPublishState.id === 'published'} type="button" onClick={this.publishAtom} className="btn btn--green btn--margin">Publish</button>
+          <button disabled={atomPublishState.id === 'published'} type="button" onClick={this.publishAtom} className="btn btn--green btn--margin">Publish</button>
 
           {this.renderTakeDownButton(atomPublishState)}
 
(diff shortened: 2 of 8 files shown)


ajwl:
[Kill the old explainer - Part 2](https://trello.com/c/Qa9XYPhG/173-kill-the-old-explainer-part-2)

ajwl:
I believe this is all I need to do. Seems to work

public/js/services/presence.js
```diff
@@ -4,7 +4,9 @@ import {logError} from '../util/logger';
 export const subscribeToPresence = (atomType, atomId) => {
   const store = getStore();
   const presenceClient = store.getState().presenceClient;
-  if (Object.keys(presenceClient).length !== 0) {
+  if (Object.keys(presenceClient).length !== 0
+      && presenceClient !== "undefined") {
+
```
ajwl: This change I have added to reduce errors in DEV environments - `presenceClient` is undefined in local because it fails to connect to workflow statuses (I think). This throws a lot of errors that clutter up the console.

akash1810: minor: an alternative to doing this check twice would be to conditionally render the component based on config. For example, this is how MAM does it - https://github.com/guardian/media-atom-maker/blob/master/public/video-ui/src/components/Header.js#L112-L120

public/js/components/AtomEdit/CustomEditors/ExplainerEditor.js
```diff
@@ -0,0 +1,24 @@
+import React, {PropTypes} from 'react';
+import {ManagedForm, ManagedField} from '../../ManagedEditor';
+import FormFieldTextInput from '../../FormFields/FormFieldTextInput';
+import FormFieldsScribeEditor from '../../FormFields/FormFieldScribeEditor';
+
+export class ExplainerEditor extends React.Component {
+
+  static propTypes = {
+    atom: PropTypes.shape({type: PropTypes.string, id: PropTypes.string}).isRequired,
```
tomrf1: is there a reason for not using `atomPropType` here?

tomrf1:
👍 great work

public/js/constants/atomData.js
```diff
@@ -72,15 +72,27 @@ export const timeline = {
...
+
+export const allAtomTypes = [cta, recipe, storyQuestions, quiz, media, qa, guide, profile, timeline, explainer];
+export const workshopEditableAtomTypes = [cta, storyQuestions, recipe, qa, guide, profile, timeline, explainer];
 export const snippetAtomTypes = [qa, guide, profile, timeline];
+export const legacyAtomTypes = [explainer];
+
 export function getNonEditableAtomTypes() {
-   return allAtomTypes.filter((atomType) => workshopEditableAtomTypes.indexOf(atomType) === -1);
+   return allAtomTypes.filter((atomType) => {
+     return !workshopEditableAtomTypes.includes(atomType) && !legacyAtomTypes.includes(atomType);
```
regiskuckaertz: nice one!

regiskuckaertz: If `LegacyAtomType(a) => WorkshopEditableAtomType(a)`, then the `&&` part is not necessary.

public/js/constants/atomData.js
```diff
@@ -72,15 +72,27 @@ export const timeline = {
...
+   });
 }
 
 export const editableNonSnippetAtomTypes =
-  workshopEditableAtomTypes.filter((atomType) => snippetAtomTypes.indexOf(atomType) === -1);
+  workshopEditableAtomTypes.filter((atomType) => {
+    return !snippetAtomTypes.includes(atomType) && !legacyAtomTypes.includes(atomType);
```
regiskuckaertz: ditto

public/js/constants/atomData.js
ajwl: Just ftr: have decided to keep `legacyAtomTypes` separate from other qualities eg being `workshopEditable` because they don't necessarily overlap. 
Hence the need for the double conditions in this filter statement

public/js/services/presence.js
ajwl: @akash1810 - thanks for raising, will look into doing this

public/js/components/AtomEdit/CustomEditors/ExplainerEditor.js
ajwl: Ah, no. Let me look at adding it in.

public/js/components/Header/EditHeader.js
```diff
@@ -12,7 +12,7 @@ class EditHeader extends React.Component {
 
   static propTypes = {
     atom: atomPropType,
-    presence: PropTypes.object,
+    presence: PropTypes.bool,
```
ajwl: This type was wrong, which it seems prevented the presence check from being used in the code

public/js/components/Header/Header.js
```diff
@@ -33,7 +34,7 @@ class Header extends React.Component {
               </Link>
             </header>
           : false}
-          {this.props.isFindPage ? <BrowseHeader /> : <EditHeader atom={this.props.atom} />}
+          {this.props.isFindPage ? <BrowseHeader /> : <EditHeader atom={this.props.atom} presence={this.props.presence}/>}
```
ajwl: Passing `presence` along to the Edit Header - so it can be used to switch off the errors found in there.

ajwl:
@regiskuckaertz - have updated the fix for the `presence` boolean, correctly passing it down from the container. Still works on code - let me know if i can merge!

PR #200, opened 2018-01-26
Migrate explainers in AW datastore

regiskuckaertz:
This is a one-off and will be reverted as soon as the migration has completed. Some explanation:

- @ajwl recently [added](https://github.com/guardian/atom-workshop/pull/198) the ability to edit explainers in atom workshop -- they used to be managed by another tool, which is going the way of the dodo
- while the UI part is done, the data store hasn't been hydrated with the necessary data, i.e. atom workshop uses a local database as a sort of index/cache/view over the whole of CAPI. 
- every action on an atom starts with a lookup into that database, but because there is currently nothing about explainers, trying to edit an explainer will fail.

In conclusion, we just need to query CAPI and populate that database with the data we get back, so that lookups won't fail and the UI loads correctly.

The migration script is a loop that queries CAPI for as long as there are pages of results and then yields to a function that inserts records in the database.

### DIFF
--- a/app/AppComponents.scala
+++ b/app/AppComponents.scala
@@ -15,9 +15,10 @@ class AppComponents(context: Context)
 
   val logger = new LogConfig
 
-  lazy val router = new Routes(httpErrorHandler, appController, healthcheckController, loginController, assets, supportController, reindex, atomActionsController)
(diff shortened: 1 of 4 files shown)


ajwl:
[Kill the old explainer - Part 2](https://trello.com/c/Qa9XYPhG/173-kill-the-old-explainer-part-2)

tomrf1:
👍 does the job!

PR #205, opened 2018-02-16
Reader questions: segregate email notifications from the publishing workflow

regiskuckaertz:
## Background

We want to automate the ability for journalists to send by email an answer they have written (in the form of a snippet) in response to votes on a readers question atom. The whole pipeline is ready and running in CODE, we just need to make sure we get the UX right in atom workshop.

## The situation right now

The current logic is such that, once a link between a question and a response has been made in atom workshop, the emails will be sent as soon as the journalist hits "publish". There is lambda bound to the Kinesis stream of CAPI updates that will filter out the proper events and kick-off the whole sending process.

### What we are changing

Journalists should choose when the emails are being sent out via an action that is separate from the "Publish" action. We don't want actions that have an impact on the readership to be triggered without them knowing, so it must result from their manifesting intent.

## The solution

In the "Notification List" right-hand column, we add a "Send" action that is displayed when a mailing list has been created (it is completely optional so far). Hitting that button will kick-off the sending of the snippet answer to subscribers.

In addition to that, there a small improvement to that right-hand column, so that it correctly reflects the current state of the system, which happens to be in either of four configurations:

1- START where no list has been created
2- CREATED where a list has been created but no email sent yet
3- IN PROGRESS where emails are being sent out
4- END where email have been sent

## TODO

- [x] Allow cross-account access to the DynamoDB table tracking emails being sent out
- [x] Update CREATED state to remove the ability to delete the list
- [x] Update state management logic to take that data into account
- [x] Update END state to show progress on the emails being sent out
- [x] Update END state to display a "Delete" action once the emails have all been sent
- [x] Fix tests
- [x] Test in CODE


app/AppComponents.scala
```diff
@@ -1,8 +1,9 @@
 import com.gu.atom.play.ReindexController
 import config.LogConfig
-import config.Config.{config, permissions}
+import config.Config.{config, permissions, dynamoDB, capiDynamoDB, capiLambdaClient}
```
hosamaly: With two `dynamoDB` configuration properties, do they have different values?
(One of the imports is unused but I'm just curious.)

regiskuckaertz: One is for access to dynamo in the capi account. If I merge them then it's so much the better.

app/db/NotificationsDB.scala
```diff
@@ -0,0 +1,38 @@
...
+
+import scala.collection.JavaConverters._
+
+class NotificationsDB(dynamoDB: AmazonDynamoDBClient) {
+  def getNotification(atomId: String, questionId: String): Either[AtomAPIError, Option[QuestionAnswers]] = {
+    val key = Seq(
+      "atomId" -> new AttributeValue().withS(atomId), 
+      "questionId" -> new AttributeValue().withS(questionId)
+    ).toMap.asJava
+    var req = new GetItemRequest(Config.notificationsDynamoTableName, key)
```
hosamaly: Does `req` need to be a `var` instead of a `val`?

app/db/NotificationsDB.scala
```diff
@@ -0,0 +1,38 @@
...
+      res
+        .flatMap(_.get("answers"))
+        .map { as =>
+          val bs = as.getL().asScala.map { a =>
+            val m = a.getM().asScala
+            for {
+              path <- m.get("answerPath").map(_.getS)
+              `type` <- m.get("answerType").map(_.getS)
+            } yield Answer(path, `type`)
+          }.flatten
```
regiskuckaertz: I used `flatten` but here it should really be `sequence` #onedayscalawillsupporttraversablewithoutcatsorscalaz

app/services/NotificationLists.scala
```diff
@@ -41,6 +47,35 @@ object NotificationLists {
...
+          var request = new InvokeRequest()
+            .withFunctionName(Config.lambdaFunctionName)
+            .withPayload(qa.asJson.toString)
+          lambdaClient.invoke(request)
+        }
+  
+        if (results.forall(_.getStatusCode == 200))
+          Either.right(atom)
+        else
+          Either.left(LambdaExecutionError(Config.lambdaFunctionName, "Unable to trigger email notifications"))
```
hosamaly: I think that it would be useful to add the error status code (and possibly the body if it's not a security issue) to the `LambdaExecutionError`. It can help with debugging.

Another curious question: Doesn't this app use logging?

app/db/NotificationsDB.scala
regiskuckaertz: Oops, forgot to change that. Good spot!

app/db/NotificationsDB.scala
```diff
@@ -0,0 +1,38 @@
...
+      "questionId" -> new AttributeValue().withS(questionId)
+    ).toMap.asJava
+    var req = new GetItemRequest(Config.notificationsDynamoTableName, key)
+    val res = Option(dynamoDB.getItem(req).getItem).map(_.asScala)
+    
+    Either.right(
+      res
+        .flatMap(_.get("answers"))
+        .map { as =>
+          val bs = as.getL().asScala.map { a =>
```
tomrf1: some of the naming could be a bit clearer here, I'm finding it quite hard to follow

public/js/components/AtomActions/NotificationList.js
```diff
@@ -6,17 +6,88 @@ class NotificationList extends Component {
...
+      ).then(
+        sent => this.setState({ notificationSent: !!sent })
+      );
+    }
+  }
+    
+  /* Everytime the atom is updated, this component receives a new
+  set of property. We must go through all the hassle of comparing
+  the current and new versions of those properties to see what
+  changed exactly. So much for being reactive. */
```
tomrf1: 😆

app/services/NotificationLists.scala
tomrf1: logging does seem quite limited in this app, though it must log the errors - https://github.com/guardian/atom-workshop/blob/021c33a0e87161621060860baf6290d5719bba0d/app/models/APIResponse.scala#L16

app/services/NotificationLists.scala
```diff
@@ -41,6 +50,42 @@ object NotificationLists {
...
+          lambdaClient.invoke(request)
+        }
+  
+        if (results.forall(_.getStatusCode == 200))
+          Either.right(atom)
+        else {
+          results
+            .foreach { r => 
+              Logger.error(s"Unable to trigger email notifications (${r.getStatusCode})")
+              Logger.error(new String(Base64.getDecoder.decode(r.getLogResult), StandardCharsets.UTF_8))
```
hosamaly: Is it an error (in this application) to get a failed response from an API?

I don't know if you have alarms that trigger every time an error is logged, but I wouldn't want to be called because an API is down for a few seconds...

app/db/NotificationsDB.scala
```diff
@@ -23,15 +23,15 @@ class NotificationsDB(dynamoDB: AmazonDynamoDBClient) {
       res
         .flatMap(_.get("answers"))
         .map { as =>
-          val bs = as.getL().asScala.map { a =>
-            val m = a.getM().asScala
+          val answers = as.getL().asScala.map { answer =>
```
tomrf1: thanks!

tomrf1:
👍 looks good!

regiskuckaertz:
I find it incredibly hard to work in this app. That kind of thing should _never_ happen—as engineers, our whole bandwidth should be devoted to solving business challenges, not having to work out how an application works. The impedance mismatch between the play framework and react is already mind-bending, but coupled with the complecting of redux and the whole event store ceremony ... I'm having nightmares just thinking about it.

There is too much going on for what is objectively a very simple application. I guess it works in a setting where there is a clear separation between server-side and client-side engineers, but that is definitely not the case here. The whole client-side stuff should really disappear, it is over-engineered just for the sake of it and is wasting our time.

app/services/NotificationLists.scala
regiskuckaertz: Good point. I'll take this as a UX improvement since it will be propagated to the UI.

PR #206, opened 2018-03-05
Simplify question/answer UX: only allow atoms as answers

regiskuckaertz:
First step in improving the UX of reader questions is to remove the ability to link an entire article as an answer to one question.

The reasoning is that, even if some questions lead to the writing of entire articles, the link between the two is not important to us in the grand scheme of things; especially, we're only interested in sending emails when the answer is a snippet.

Following that change, I will simplify the model and update our back-end to reflect that.


tomrf1:
👍

PR #207, opened 2018-03-07
Providing answers to reader questions: improve UX

regiskuckaertz:
_Asking the good Sir @shaundillon to please have a look and advise on how I could perhaps reuse the existing infrastructure instead of writing my own._

Right now if people want to link a snippet to a question, they have to manually copy-paste the URL and browser from one page to another inside Atom Workshop.

screen shot 2018-03-07 at 14 50 45

This is a time-consuming process but is also error-prone. To remediate this problem we replace the input field with a search facility that lets users type in keywords:

screen shot 2018-03-07 at 14 41 38

A list of suggestions appear as soon as the user has entered at least three characters and CAPI provides results:

untitled

Selected snippets are displayed underneath in a format that shows both the snippet title as well as its ID:

screen shot 2018-03-07 at 14 42 07

Because the Thrift definition for answers only contain the ID of the snippet, we query CAPI when the component is mounted to retrieve the title of each atom. This ensures the user always knows what he's looking at.

## A short explanation on the inner workings and a confidence

OK let's start with the latter. I don't know how to use Redux 😓 I know it is some sort of FSM, however I don't know this ins and outs. In my case, I needed something that reacted to user inputs, queried CAPI at the right moment, displayed results and let the user select one. So I built my own little state machine (`machine`). It is simplicity itself:

- keys are states
- values are dictionaries of transitions to new states

The `transition` function updates the state and performs the required action accordingly. Each action will update the state of the component so that React can work its magic and render the updated result.

It is trivial to generalise that process to any component, and I suppose that's what Redux does. Or does it?

## Updated design

I also took this opportunity to remove a lot of visual complexity in the array component. True, atom workshop could use some design love, but I believe with these changes it is a bit easier on the eye. 

Replacing the answer input also allowed me to remove one layer of grouping, so there is no "Move up/down" buttons anymore for answers—they were useless anyway.


public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
```diff
@@ -27,21 +39,43 @@ export class StoryQuestionsQuestion extends React.Component {
...
-    const answersCount = (this.props.fieldValue && this.props.fieldValue.answers) ? this.props.fieldValue.answers.length : 0;
+  onSelect = (snippet) => {
+    this.setState((oldState) => Object.assign({}, oldState, { answers: oldState.answers.concat({
+      title: snippet.title,
+      answerId: `atom/${snippet.atomType}/${snippet.id}`, 
+      answerType: 'ATOM' 
+    })}));
+  }
 
+  onDelete = (i) => {
```
hosamaly: I find it generally safer to rely on an `answerId` instead of an index when deleting stuff. The user might click the button twice quickly*, and then you might receive two delete events with the same index. The code would then delete two items even though the user only wanted to delete one.

\* _Theoritically, this shouldn't happen unless there are worker threads, but I don't know if React uses / will use them._

regiskuckaertz: Yeah there are probably UX pitfalls all over the place. I need to iron out an elegant way to polish that, hopefully without muddying the logic of my components. This kind of thing should be handled at a lower level and not surface in your business logic. (no worker thread)

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
```diff
@@ -27,21 +39,43 @@ export class StoryQuestionsQuestion extends React.Component {
     this.props.onUpdateField(questionWithId);
   }
 
-  render () {
-    const answersCount = (this.props.fieldValue && this.props.fieldValue.answers) ? this.props.fieldValue.answers.length : 0;
+  onSelect = (snippet) => {
```
tomrf1: I can't understand how the answers in the props get updated if we're only maintaining `state.answers`

regiskuckaertz: I forgot to send the last two commits 😅

tomrf1: oh, it looks like the code has updated in my browser since I wrote this comment. I can see how it works now!

public/js/components/FormFields/SearchFields/SearchSuggestions.js
```diff
@@ -0,0 +1,156 @@
...
+        this.transition('RESOLVE');
+      })
+      .catch(error => {
+        this.setState({ error });
+        this.transition('REJECT');
+      });
+  }
+
+  displayError = () => {
+
```
tomrf1: should something happen here? Not sure how we should communicate this to the user, I did something very basic [here](https://github.com/guardian/atom-workshop/blob/b2001d1484c8cd99768c0e9a7d5bd4d9a75ed656/public/js/components/CommonsDivisions/CommonsDivision.js#L169)

tomrf1:
👍 looks good!

public/js/components/FormFields/SearchFields/SearchSuggestions.js
regiskuckaertz: oh yeah you're right, totally forgot that 😆 I'm hoping the red band component can be used here. let's see.

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
```diff
@@ -1,9 +1,13 @@
 import React, { PropTypes } from 'react';
 import {ManagedForm, ManagedField} from '../../../ManagedEditor';
 import FormFieldTextInput from '../../../FormFields/FormFieldTextInput';
-import FormFieldArrayWrapper from '../../../FormFields/FormFieldArrayWrapper';
-import {StoryQuestionsAnswer} from './Answer';
+import SearchSuggestions from '../../../FormFields/SearchFields/SearchSuggestions';
 import uuidv4 from 'uuid/v4';
+import { fetchCapiAtom } from '../../../../services/capi';
+
+const filters = {
```
shaundillon: I'd maybe abstract this out into the `consts` folder, not that it's any more robust, but it makes it easier to reuse if needed.

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
```diff
@@ -19,29 +23,93 @@ export class StoryQuestionsQuestion extends React.Component {
     onFormErrorsUpdate: PropTypes.func
   };
 
+  constructor(props) {
+    super(props);
```
shaundillon: Constructor blocks aren't required because of the `transform-class-properties` babel plugin. Props are already applied as `this.props` and state (or anything for that matter) can be set as a called properly by simply using:
```
state = {
...
}
```

inside the class declaration.

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
```diff
@@ -19,29 +23,93 @@ export class StoryQuestionsQuestion extends React.Component {
...
+    this.state = {
+      answers: props.fieldValue && props.fieldValue.answers && props.fieldValue.answers.length ?
```
shaundillon: As lodash is included, you can make this a lot more terse using `_.get()`

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
```diff
@@ -19,29 +23,93 @@ export class StoryQuestionsQuestion extends React.Component {
...
     });
 
     this.props.onUpdateField(questionWithId);
   }
 
-  render () {
-    const answersCount = (this.props.fieldValue && this.props.fieldValue.answers) ? this.props.fieldValue.answers.length : 0;
+  onSelect = (snippet) => {
+    this.setState(
+      (oldState) => Object.assign({}, oldState, { answers: oldState.answers.concat({
```
shaundillon: We don't need to update the whole state here, sending a new answers block is enough:
``` js
(oldState) => {
  answers: oldState.answers.concat(...etc);
}
```

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
```diff
@@ -19,29 +23,93 @@ export class StoryQuestionsQuestion extends React.Component {
...
+          questionText: this.props.fieldValue.questionText,
+          answers: this.state.answers
+        });
+      }
+    );
+  }
+  
+  onDelete = (i) => {
+    this.setState(
+      (oldState) => Object.assign({}, oldState, {
```
shaundillon: As above

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
```diff
@@ -19,29 +23,93 @@ export class StoryQuestionsQuestion extends React.Component {
...
-              <StoryQuestionsAnswer/>
-            </FormFieldArrayWrapper>
-          </ManagedField>
         </ManagedForm>
+        <div className="atom__answers__header">
+          <h5>Answers</h5>
+          <SearchSuggestions fieldPlaceholder="Search for snippets" filters={filters} onSelect={this.onSelect} />
+        </div>
+        <ul className="atom__answers">
+          {this.state.answers.map((answer, i) =>
```
shaundillon: The only way I can see this could reuse other components is to use the `FormFieldArrayWrapper` but it's a bit specialised, I think it'd be too much effort.

public/js/components/FormFields/SearchFields/SearchSuggestions.js
```diff
@@ -0,0 +1,156 @@
...
+      KEY: 'type',
+      ESC: 'idle',
+      SEL: 'idle'
+    },
+    error: {
+      KEY: 'type'
+    }
+  };
+  
+  constructor(props) {
```
shaundillon: As above, these constructor blocks aren't needed.

public/styles/components/_search-suggestions.scss
```diff
@@ -0,0 +1,36 @@
...
+    position: absolute;
+    top: 100%;
+    background: white;
+    list-style: none;
+    margin: 0;
+    padding: 0;
+}
+
+.search-suggestions > ul > li {
+    padding: .75em 1em;
```
shaundillon: `em` is a little flimsy as it's relative to the parent's font size. I'd use `rem`

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
hosamaly: Would this cause us to lose any checks done by the linter?
(I don't know if the linter can even check this; excuse my ignorance!)
```

**Comment by regiskuckaertz** [comment]
Great! Thanks for your comments, I'm happy to pair with you on the Redux refactoring next week so we can abstract that away.

File: public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js (hunk 6 above)
Comment (hosamaly) [reply public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js]: An alternative (for the fun of code golf):
```javascript
answers: (props.fieldValue || {}).answers || []
```

shaundillon: @hosamaly lodash would be `_.get(props.fieldValue, 'answers')` I think. Not sure about the linter, depends on rules but it shouldn't, no

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
regiskuckaertz: yeah no, let's not do that 😅 I'm happy to refactor that module (the array wrapper one) another time, there are some abstraction leaks which make it impractical to work with

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
```diff
@@ -19,29 +25,88 @@ export class StoryQuestionsQuestion extends React.Component {
     onFormErrorsUpdate: PropTypes.func
   };
 
+  state = {
+    answers: _get(this.props, 'fieldValue.answers', [])
```
regiskuckaertz: That expression returns `undefined`, despite there being a default value. Any idea why, @shaundillon ?

shaundillon: I think there's a typo, it should be `_.get`

regiskuckaertz: I'm importing `'lodash/fp/get'` as done in other modules, isn't that the preferred method?

public/js/components/FormFields/SearchFields/SearchSuggestions.js
```diff
@@ -12,6 +27,8 @@ export default class SearchSuggestions extends React.Component {
   };
 
   initialState = 'idle';
+
+  store = createStore(this.transition, this.initialState);
```
shaundillon: We could use the existing store. There's a block at the bottom of some of the components that create a 'connected component', that is, a component with access to the redux store on props. You can use this to pass action creators into a component and call changes to the store, which are passed back down as props.

If you're free tomorrow, I can come pair on this with you?

regiskuckaertz: Yep that'd be most welcome 😽

regiskuckaertz:
@shaundillon let me know if you're happy with the above changes

PR #208, opened 2018-03-13
Enable rich text toolbar on text inputs for atoms

ajwl:
This will allow the text editing toolbar to be available on atom text input fields.
Allowing bold, italic and links to be used. 

It was requested by UX, we will inform Central production. 

Previously adding text to an atom looked like this:

screen shot 2018-03-13 at 16 38 04

Now it looks like:
screen shot 2018-03-14 at 10 43 27


ajwl:
Have told Central Production.

PR #209, opened 2018-03-14
Change snippet descriptions

regiskuckaertz:


LATaylor-guardian:
👍

public/js/constants/atomData.js
```diff
@@ -51,19 +51,19 @@ export const media = {
 export const qa = {
   type: "qanda",
   fullName: "Q & A",
-  description: "Provide readers with the answers to their questions"
+  description: "A <b>single</b> question with a <b>single</b> answer"
```
hosamaly: Is using `<b>` tags generally standard here? Is there any preference against `<em>` or `<strong>`?

regiskuckaertz: the `B` vs `STRONG` dichotomy dates back to when `B` only had a visual meaning, but in the last few years it (and all the others like `I`, `S`, etc.) have been revived in [the HTML5 spec](https://www.w3.org/TR/html52/textlevel-semantics.html#the-b-element):

> The `b` element represents a span of text to which attention is being drawn for utilitarian purposes without conveying any extra importance and with no implication of an alternate voice or mood

hosamaly: I didn't know that. Thanks for the link!

PR #210, opened 2018-03-22
Allow multiple search suggestions

regiskuckaertz:
The global state currently holds properties for a single instance of `SearchSuggestions`, resulting in _all_ instances being hydrated with the same properties. If the users types something in one search field, all fields will display the same value and results. This is fun.

I've now made room for more that one instance. Each instance has an ID that is randomly generated—let's not be anal about that—and this ID is propagated into the global state.

Also a couple of display issues such as spacing and stuff like that.

### DIFF
--- a/public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
+++ b/public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
@@ -29,6 +29,12 @@ export class StoryQuestionsQuestion extends React.Component {
       this.props.fieldValue.answers : []
   }
 
+  searchSuggestionsId = this.genKey()
+
+  genKey() {
+    return 'xxxxxxx'.replace(/x/g, () => (Math.random() * 36 | 0).toString(36));
+  }
+
   componentDidMount() {
     this.state.answers.forEach((answer, i) => {
     // eslint-disable-next-line
@@ -94,7 +100,7 @@ export class StoryQuestionsQuestion extends React.Component {
         </ManagedForm>
         <div className="atom__answers__header">
           <h5>Answers</h5>
-          <SearchSuggestions fieldPlaceholder="Search for snippets" filters={filters} onSelect={this.onSelect} />
+          <SearchSuggestions id={this.searchSuggestionsId} fieldPlaceholder="Search for snippets" filters={filters} onSelect={this.onSelect} />
         </div>
         <ul className="atom__answers">
           {this.state.answers.map((answer, i) => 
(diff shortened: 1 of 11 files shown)


public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
```diff
@@ -94,7 +100,7 @@ export class StoryQuestionsQuestion extends React.Component {
         </ManagedForm>
```
ajwl: It adds clutter but would be good to give a prompt to the user that the answer has to be another atom
`<small> Answer the question with an atom, select one below </small>`

ajwl:
I don't know what the user flow with editing published Answers is... but the SearchSuggestions doesn't seem to work, after the atom has been published.
Maybe these would not be edited after publication? 

screen shot 2018-03-22 at 13 04 43

ajwl:
Would be ideal if search worked on partial words instead of only whole words: 
Ie: "Zeek" doesn't work, but "Zeeks" does. Ideally partial words would match to the name: 

screen shot 2018-03-22 at 13 13 23

screen shot 2018-03-22 at 13 13 30

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
regiskuckaertz: That's a good idea, thanks 👍

regiskuckaertz:
Yeah looking into that now.

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
```diff
@@ -29,6 +29,12 @@ export class StoryQuestionsQuestion extends React.Component {
       this.props.fieldValue.answers : []
   }
 
+  searchSuggestionsId = this.genKey()
+
+  genKey() {
+    return 'xxxxxxx'.replace(/x/g, () => (Math.random() * 36 | 0).toString(36));
```
hosamaly: What's the value of `| 0`?

ajwl:
@regiskuckaertz hm. Yes. I think users will be expecting the search functionality they get with eg Google where each letter triggers a search. 
However if this involves lots of other refactoring, then your suggestion of a prompt could be a good part-way solution.

screen shot 2018-03-22 at 14 35 56

However,

public/js/components/AtomEdit/CustomEditors/StoryQuestionFields/Question.js
regiskuckaertz: Truncation of the decimal part.

hosamaly: Thanks. I wonder whether the whole method can be defined as `Math.random().toString(36)`. It seems to have a similar result in both Chrome and Firefox.

regiskuckaertz: Definitely 👍

ajwl:
👍  looks good to me

PR #212, opened 2018-05-30
Simplify target creation

tomrf1:
We've had a few cases now of people struggling to create targets for their atoms.
It's consistently because they don't realise they need to click the 'Create Suggestion' button at the end. The UI is quite confusing and should really be rebuilt one day.

For now, as a quick fix I have tried to make it a bit clearer:
1. The 'Create Suggestion' button is disabled until a tag is added, then the button becomes active and green.
2. Removed the 'active until' date stuff, as Chris Moran says they don't need it and it takes up loads of space in the middle of the component. Instead it just sets the (mandatory) date to 50 years from now

### Before:
picture 5

### After:
picture 4


ajwl:
Closing as I believe this is now out of date

PR #213, opened 2018-06-15
Fix delete button

mbarton:
Ronseal. I have refactored the server-side code to issue a takedown followed by delete for published content and simply a delete for unpublished. Finally I changed the frontend to redirect to the root, where previously it cleared the state and then barfed React render errors all over the place.

We should probably disallow deleting the atom until all usages have been removed.

6cf8d28bd540fa4dd0b265b92e3b3ff9

### DIFF
--- a/public/js/actions/AtomActions/deleteAtom.js
+++ b/public/js/actions/AtomActions/deleteAtom.js
@@ -31,7 +31,7 @@ function errorDeletingAtom(error) {
 export function deleteAtom(atom) {
   return dispatch => {
     dispatch(requestAtomDelete(atom));
-    return AtomsApi.DeleteAtom(atom)
+    return AtomsApi.deleteAtom(atom)
         .then(res => res.json())
         .then(atom => {
           dispatch(receiveAtomDelete(atom));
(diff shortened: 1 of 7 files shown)


public/js/actions/AtomActions/deleteAtom.js
```diff
@@ -31,11 +22,13 @@ function errorDeletingAtom(error) {
...
-    return AtomsApi.DeleteAtom(atom)
+
+    return AtomsApi.deleteAtom(atom)
         .then(res => res.json())
-        .then(atom => {
-          dispatch(receiveAtomDelete(atom));
-        })
-        .catch(error => dispatch(errorDeletingAtom(error)));
+        .then(() =>
+          window.location.href = window.location.origin + '/'
```
shaundillon: Any reason for this over using the history from react router?

mbarton: Ah yes sorry, should've left a comment.

The only reason is laziness, `BrowserHistory.push` didn't work when I tried it locally. I suspect it's the same reason why the back button is broken at the moment.

mbarton: @shaundillon and I have discussed IRL and I will raise a separate PR to fix it

PR #214, opened 2018-06-18
Fix back button

mbarton:
The back button was busted since it replaced the query parameters with an empty object rather than the default (leading to rendering errors in child components).

Now it doesn't! Success!


tomrf1:
💯

PR #216, opened 2018-10-08
Add chart atom to Atom Workshop

susiecoleman:
Allow a chart atom to be created in atom workshop. Iframes in the [basichartool](https://visuals.gutools.co.uk/basichartool/) to allow editing.

### DIFF
--- a/public/js/components/AtomEdit/CustomEditors/ChartEditor.js
+++ b/public/js/components/AtomEdit/CustomEditors/ChartEditor.js
@@ -1,6 +1,7 @@
 import React, { PropTypes } from 'react';
 import Modal from '../../Utilities/Modal';
 import {atomPropType} from '../../../constants/atomPropType';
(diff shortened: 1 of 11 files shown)


public/js/components/AtomEdit/CustomEditors/ChartEditor.js
```diff
@@ -0,0 +1,71 @@
...
+  };
+
+  closeModal = () => {
+    this.setState({ modalOpen: false });
+    window.removeEventListener('message', this.onMessage, false);
+  };
+
+  openModal = () => {
+    this.setState({ modalOpen: true });
+    window.addEventListener('message', this.onMessage, false);
```
akash1810: `this.onMessage` doesn't exist?

susiecoleman: Whoops, copy and paste fail. It is now there

PR #219, opened 2018-10-15
show a message if not authed against Visuals Chart Tool

akash1810:
Sadly we can't use the auth flow of the Visuals Chart Tool within the iframe as Google Auth complains `Refused to display ... in a frame because it set 'X-Frame-Options' to 'sameorigin'.`

This is a cheat to provide a slightly better user experience when not authed.

# Before
before

# After
after


akash1810:
[Fix auth log in when iframing chart tool into atom workshop](https://trello.com/c/blxiJRSu/25-fix-auth-log-in-when-iframing-chart-tool-into-atom-workshop)

PR #221, opened 2018-10-16
make request to visuals with a cookie

akash1810:
We're making a call to visuals and checking the response code as a horrible way to check if we're authed.

It would help if we [send our cookie along with the request](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#Sending_a_request_with_credentials_included)...

relates to #219


akash1810:
So, this doesn't work in PROD because of cors and all that wonderful stuff...

https://github.com/guardian/atom-workshop/pull/222

PR #225, opened 2019-01-14
AUDIO ATOM: fix data structure 

ajwl:
Audio Atom data fields were saving on the root of the object, not inside the `data` property.


regiskuckaertz:
I mean, locally is fine too 👍

ajwl:
yes indeed tested in code.
Thanks!

PR #231, opened 2019-04-18
Bump thrift-related dependencies

regiskuckaertz:
I've done a tag search and verified that updating an atom made it through to CAPI in code. Is there any other scenario I should test?


regiskuckaertz:
Yep, done. Thanks 👍

PR #233, opened 2019-04-29
Add a temporary route for reindexing explainers

regiskuckaertz:
As part of the move to the new atoms index, all existing atoms must be backfilled. This is proving difficult for explainers in particular, as we decommissioned them a while ago: the authoring app, in particular, is gone. To avoid resurrecting it, it is much simpler to use atom workshop instead, which already has nice interfaces with DynamoDB and Kinesis.

There are not many explainer atoms (~300 on PROD), so I've kept the code to the bare minimum.

- [x] Tested on CODE
- [x] Update floodgate to use the new route on CODE
- [x] Update floodgate to use the new route on PROD

### DIFF
--- a/app/controllers/ExplainerReindexController.scala
+++ b/app/controllers/ExplainerReindexController.scala
@@ -10,6 +10,7 @@ import com.gu.contentatom.thrift.{ContentAtomEvent, EventType}
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.Try
 import play.api.libs.json.Json
+import play.api.Logger
 
 class ExplainerReindexController(
   val wsClient: WSClient,
@@ -56,6 +57,8 @@ class ExplainerReindexController(
     Ok(Json.parse(s"""{ "status": "completed", "documentsIndexed": $result, "documentsExpected": $result }"""))
 
   private val displayError: PartialFunction[Throwable, Result] = {
-    case x: Throwable => InternalServerError(Json.parse(s"""{ "status": "failed", "documentsIndexed": 0, "documentsExpected": 0 }"""))
+    case x: Throwable => 
+      Logger.error("Something went wrong", x)
+      InternalServerError(Json.parse(s"""{ "status": "failed", "documentsIndexed": 0, "documentsExpected": 0 }"""))
   }
 }
--- a/app/controllers/ExplainerReindexController.scala
+++ b/app/controllers/ExplainerReindexController.scala
@@ -5,7 +5,7 @@ import com.gu.atom.publish.{AtomReindexer, PreviewAtomReindexer, PublishedAtomRe
 import db.ExplainerDBAPI
 import play.api.Configuration
 import play.api.libs.ws.WSClient
-import play.api.mvc.{ Action, AnyContent, Controller, Result }
+import play.api.mvc.{ Action, ActionBuilder, AnyContent, Controller, Result, Request }
 import com.gu.contentatom.thrift.{ContentAtomEvent, EventType}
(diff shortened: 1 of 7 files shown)


regiskuckaertz:
Integration with floodgate fully tested 🎉

app/controllers/ExplainerReindexController.scala
```diff
@@ -0,0 +1,79 @@
...
+  private def updateState(isPreview: Boolean): PartialFunction[Try[Int], _] = {
+    case Success(x) => if (isPreview) lastPreview = x else lastPublished = x
+  }
+
+  private def displayResult(result: Int): Result =
+    Ok(Json.parse(s"""{ "status": "completed", "documentsIndexed": $result, "documentsExpected": $result }"""))
+
+  private val displayError: PartialFunction[Throwable, Result] = {
+    case x: Throwable => 
+      Logger.error("Something went wrong", x)
```
mbarton: Let's be a bit more descriptive about what action we were trying to perform, like reindexing preview or published

app/controllers/ExplainerReindexController.scala
```diff
@@ -0,0 +1,79 @@
...
+      case "published" => displayResult(lastPublished)
+      case x => BadRequest(s"$x is not a valid stack identifier")
+    }
+  }
+
+  private def run(datastore: DynamoDataStore, reindexer: AtomReindexer): Future[Int] = {
+    explainerDB.listAtoms(datastore).fold(
+      apiError => Future.failed(new RuntimeException(apiError.msg)),
+      { atoms =>
+        val timestamp = System.nanoTime
```
mbarton: Not sure, do atoms have nano time precision or millis?

app/controllers/ExplainerReindexController.scala
```diff
@@ -0,0 +1,79 @@
...
+  val wsClient: WSClient,
+  explainerDB: ExplainerDBAPI,
+  previewDataStore: DynamoDataStore,
+  publishedDataStore: DynamoDataStore,
+  previewReindexer: PreviewAtomReindexer,
+  publishedReindexer: PublishedAtomReindexer,
+  config: Configuration
+)(implicit ec: ExecutionContext) extends Controller with PanDomainAuthActions {
+
+  var lastPublished: Int = 0
```
mbarton: Will these guys get set back to zero on completion by the library?

app/controllers/ExplainerReindexController.scala
```diff
@@ -0,0 +1,79 @@
...
+  def status(stack: String): Action[AnyContent] = ApiKeyAction { req =>
+    stack match {
+      case "preview" => displayResult(lastPreview)
+      case "published" => displayResult(lastPublished)
+      case x => BadRequest(s"$x is not a valid stack identifier")
+    }
+  }
+
+  private def run(datastore: DynamoDataStore, reindexer: AtomReindexer): Future[Int] = {
+    explainerDB.listAtoms(datastore).fold(
```
akash1810: might be worth checking the provisioned iops on that table and bump it if needed during this operation

regiskuckaertz: good call! I checked and it's actually over-provisioned. This will only get called once in a lifetime and the table will never grow afaik, so even a `scan` is harmless.

app/controllers/ExplainerReindexController.scala
regiskuckaertz: Nope, that's not necessary (the reindex controller for regular atoms does not either)

regiskuckaertz:
All good now?

PR #235, opened 2019-05-10
Use `Tag#internalName` for autocomplete in line with Composer

regiskuckaertz:
This is a long-standing request by CP that AW uses the tags' internal name for type-ahead recommendations. In this PR, I made two changes:

- uses the new `internal-name` filter for querying tags. I'm also forcing the request on the internal tier, though I'm not sure this is necessary for preview but I couldn't find anything related in Concierge
- prevents a request to be done when the search text is empty

Screenshot 2019-05-10 at 10 56 22

Additionally I fixed a local dev bug where the explainer reindex controller couldn't start because of some missing configuration.

### DIFF
--- a/app/config/Config.scala
+++ b/app/config/Config.scala
@@ -61,8 +61,8 @@ object Config extends AwsInstanceTags {
 
   val previewDynamoTableName = config.getString("aws.dynamo.preview.tableName")
   val publishedDynamoTableName = config.getString("aws.dynamo.live.tableName")
-  val explainerPreviewDynamoTableName = getOptionalProperty("aws.dynamo.explainers.preview.tableName", config.getString).getOrElse("")
-  val explainerPublishedDynamoTableName = getOptionalProperty("aws.dynamo.explainers.live.tableName", config.getString).getOrElse("")
+  val explainerPreviewDynamoTableName = getOptionalProperty("aws.dynamo.explainers.preview.tableName", config.getString).getOrElse("explain-maker-preview-DEV")
+  val explainerPublishedDynamoTableName = getOptionalProperty("aws.dynamo.explainers.live.tableName", config.getString).getOrElse("explain-maker-live-DEV")
   val notificationsDynamoTableName = config.getString("aws.dynamo.notifications.tableName")
 
   val gridUrl = config.getString("grid.url")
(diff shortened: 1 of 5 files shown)


public/js/components/FormFields/FormFieldTagPicker.js
```diff
@@ -39,20 +39,27 @@ export default class FormFieldTagPicker extends React.Component {
 
     const searchText = e.target.value;
 
-    this.setState({
-      searchText: searchText
-    });
-
-    searchTags(searchText, this.props.tagType).then((results) => {
+    if (searchText.length > 0) {
```
JustinPinner: Might be worth making this a minimum number of chars before the search starts. I've seen three used in other contexts but that'd depend on whether we have any one or two character tags I guess (but hope that's not the case)

regiskuckaertz: Great point! Is Composer doing that too? I'll have a look

regiskuckaertz: It kicks off with one character in Composer, so I think it's best to align AW with it. Though I think you're right 😞

regiskuckaertz:
Sorry, last minute change. I noticed that we were querying the CAPI preview stack, but it is not possible to reindex it using floodgate (it is disabled for some reason, I'll investigate later). What that means is that `internalName` is almost always empty there.

I don't think it makes sense to search for tags in the preview stack anyway, so in the last change I added a route to query the live CAPI stack instead. Currently only used for tag searches.

cc @akash1810 @itsibitzi

app/config/Config.scala
```diff
@@ -83,6 +83,8 @@ object Config extends AwsInstanceTags {
   val capiPreviewIAMUrl = config.getString("capi.previewIAMUrl")
   val capiLiveUrl = config.getString("capi.liveUrl")
 
+  val capiApiKey = getOptionalProperty("capi.apiKey", config.getString).getOrElse("atom-workshop-DEV")
```
akash1810: are we happy with this key being in source control?

regiskuckaertz: no 😭 sorry about that, will fix.

regiskuckaertz: actually, no. I've demoted it to a developer key and added an internal key to AW CODE.

PR #236, opened 2019-05-10
Change word limit in guide atoms

walaura:
We only have the new style error messages in QnA snippets. This brings the new messaging over to the guide snippets and a followup pr will bring it to the other 2

Screenshot 2019-05-10 at 15 17 35

### DIFF
--- a/public/js/components/AtomEdit/CustomEditors/GuideEditor.js
+++ b/public/js/components/AtomEdit/CustomEditors/GuideEditor.js
@@ -1,36 +1,42 @@
 import React, { PropTypes } from 'react';
 import FormFieldImageSelect from '../../FormFields/FormFieldImageSelect';
 import FormFieldArrayWrapper from '../../FormFields/FormFieldArrayWrapper';
-import {GuideItem} from './GuideFields/GuideItem';
-import {ManagedField, ManagedForm} from '../../ManagedEditor';
-import {atomPropType} from '../../../constants/atomPropType';
-import {checkItemsUnderWordCount} from '../../../util/validators';
+import { GuideItem } from './GuideFields/GuideItem';
+import { ManagedField, ManagedForm } from '../../ManagedEditor';
+import { atomPropType } from '../../../constants/atomPropType';
 
 export class GuideEditor extends React.Component {
+	static propTypes = {
+		atom: atomPropType.isRequired,
+		onUpdate: PropTypes.func.isRequired,
+		onFormErrorsUpdate: PropTypes.func,
+		config: PropTypes.shape({
+			gridUrl: PropTypes.string.isRequired,
+		}).isRequired,
+	};
 
-  static propTypes = {
-    atom: atomPropType.isRequired,
-    onUpdate: PropTypes.func.isRequired,
-    onFormErrorsUpdate: PropTypes.func,
-    config: PropTypes.shape({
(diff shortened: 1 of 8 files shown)


public/js/util/validators.js
```diff
@@ -1,44 +1,51 @@
...
+		var wordCount = values.map(function(itemData) {
+			var strippedText = stripHtml(itemData.body);
+			return strippedText.split(' ').length;
+		});
+
+		var totalCount = wordCount.reduce(function(total, num) {
+			return total + num;
+		});
+
+		if (totalCount <= wordLimits.default) {
```
regiskuckaertz: This test is probably wrong, the word limit should be passed as a parameter I guess?

public/js/util/validators.js
```diff
@@ -1,44 +1,51 @@
...
-
-        if (totalCount <= WordLimit) {
-            return Promise.resolve(true);
-        } else {
-            const error = new FieldError('too-long', `You\'ve reached the word limit on this atom type. Please edit your items to less than a total of ${WordLimit} words.`);
-            return Promise.resolve(error);
-        }
-
-    }
+export const checkItemsUnderWordCount = values => {
```
regiskuckaertz: The function is never called anywhere 😅

public/js/util/validators.js
walaura: this was the behavior already in place and I totally agree with you for this but since we wanna not treat word limits as form errors moving forward I'd wait and see if it makes sense to put effort here

public/js/util/validators.js
walaura: happy to remove it, I have no idea what it does!

walaura: hey this is being called in `TimelineEditor`!

regiskuckaertz: It is the function that is responsible for displaying the message. You've actually removed it from the guide editor: was that intended? If yes sorry, I don't see what is picking up the word count then?

public/js/components/AtomEdit/CustomEditors/GuideEditor.js
```diff
@@ -1,30 +1,36 @@
...
+          formName="guideEditor"
+        >
+          <ManagedField
+            fieldLocation="data.guide.guideImage"
+            name="Guide Image"
+          >
+            <FormFieldImageSelect gridUrl={this.props.config.gridUrl} />
           </ManagedField>
-          <ManagedField fieldLocation="data.guide.items" name="Items" customValidation={[checkItemsUnderWordCount]}>
+          <ManagedField fieldLocation="data.guide.items" name="Items">
```
regiskuckaertz: Here. Is that correct?

public/js/util/validators.js
regiskuckaertz: ok, it's in the scribe editor now. Let's leave it there for now, but remember to clean it up when you're done moving the other two snippet types 🤗

public/js/components/AtomEdit/CustomEditors/GuideEditor.js
walaura: yes! instead of attaching a validator to check for the word limit to error the form we are using a prop in the textfield (This was already built before I joined, super handy!) to show a notice if a word limit is hit. The reason for this is that word limits aren't errors but more of a suggestion!

PR #238, opened 2019-05-13
Disable publish/save button on form errors

walaura:
This is prep work for making the tags mandatory. Now the button disables itself if a form contains errors. The tags not being there will be an error and this UI allows us to actually disable the button

Screenshot 2019-05-13 at 15 00 16

This disabling _may_ cause havoc – I've had a look at our errors and most seem to be replicated on the backend. let's merge it in the morning and revert it if it breaks any workflows?


public/js/components/Header/EditHeader.js
```diff
@@ -1,128 +1,161 @@
...
+          <PresenceIndicator presence={this.props.presence} />
+        ) : (
+          false
+        )}
+
+        <HoverExpander
+          proxy={
+            <button
+              disabled={
+                atomPublishState.id === 'published' ||
```
regiskuckaertz: Let's move that expression in a function for clarity:
```js
disabled={this.isPublishingDisabled()}
```

public/styles/components/_hover-expander.scss
```diff
@@ -0,0 +1,26 @@
...
+.hover-expander__trigger {
+  display: flex;
+}
+
+.hover-expander__content {
+  position: absolute;
+  top: 100%;
+  right: 0;
+  background: #fff;
+  min-width: 200px;
```
regiskuckaertz: Should there be a max-width too?

public/styles/components/_hover-expander.scss
```diff
@@ -0,0 +1,26 @@
+.hover-expander {
+  position: relative;
```
regiskuckaertz: Let's add
```suggestion
  position: relative;
  contain: size style layout;
```

public/styles/main.scss
```diff
@@ -4,44 +4,24 @@
...
-'components/atom-embed',
-'components/atom-actions',
-'components/workflow',
-'components/targeting',
-'components/suggestions',
-'components/commons-divisions',
-'components/search-suggestions',
-'components/chart',
-'components/audio';
+@import 'components/buttons', 'components/forms', 'components/tables',
```
regiskuckaertz: Hmm. This is not ideal for readability. How about one `@import` per line?

public/js/components/Utilities/ShowAllErrors.js
```diff
@@ -0,0 +1,33 @@
+import React, { PropTypes } from 'react';
+import flattenFormErrors from '../../util/flattenFormErrors';
+
+class ShowAllErrors extends React.Component {
```
regiskuckaertz: It looks like this is just a glorified list. How about factoring out the error stuff, so that it becomes more reusable? (and you can avoid having to deep-flatten twice).

walaura: good shout!

public/styles/main.scss
walaura: huh it actually used to be in a single line before? i think i 'made' it multiline on a different pr accidentally using prettier. Can make it multiline again, I preffer it

public/styles/components/_hover-expander.scss
walaura: neat, yeah

public/styles/components/_hover-expander.scss
walaura: not really a need at this point if you ask me?

public/styles/main.scss
regiskuckaertz: Yes, multiline.

public/styles/components/_hover-expander.scss
walaura: hey, I had to take out `size`. This is because the element doesn't have an intrinsic size (it depends on the children) so it was rendering at `0x0px`. This sounds like expected behaviour to me so I just removed it

PR #239, opened 2019-05-13
Display tag's internal name

regiskuckaertz:
Use tags' internal name instead of web title when displaying autocomplete results, if it is available.


regiskuckaertz:
[Atom Workshop: display tags' internal name](https://trello.com/c/JtfPFb61/704-atom-workshop-display-tags-internal-name)

PR #240, opened 2019-05-17
Make snippets require tagging

walaura:
BOOSH

*As an editorial user, I want to be able to find existing atoms in Composer, so that I can efficiently re-use existing content.*

**Issue**
Many atoms are being created without tags. The result is that they don't appear on Composer / Proposer. 

Screenshot 2019-05-17 at 13 19 52
Screenshot 2019-05-17 at 13 20 06

### DIFF
--- a/public/js/components/AtomEmbed/AtomEmbed.js
+++ b/public/js/components/AtomEmbed/AtomEmbed.js
@@ -3,6 +3,7 @@ import { atomPropType } from '../../constants/atomPropType.js';
 import copy from 'copy-to-clipboard';
 import { Link } from 'react-router';
 import Workflow from '../Workflow/Workflow';
+import { doesAtomTypeRequireTagging } from '../../constants/atomData';
 
 import CurrentTargets from './CurrentTargets';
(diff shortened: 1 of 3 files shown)


public/js/components/AtomEmbed/AtomEmbed.js
```diff
@@ -66,26 +70,45 @@ class AtomEmbed extends React.Component {
...
+              config={this.props.config}
+              workflow={this.props.workflow}
+              workflowActions={this.props.workflowActions}
+            />
           </div>
           <div className="form__row">
-            <h3 className="form__subheading">Suggest This Atom</h3>
+            <h3 className="form__subheading">
+              Tags
+              {doesAtomTypeRequireTagging(this.props.atom.atomType) && ' *'}
```
regiskuckaertz: won't this display `false` if the test is false?

walaura: Nope! Jsx hides booleans to allow this syntax

PR #241, opened 2019-05-20
Ensure a duration is present when importing an audio episode

JustinPinner:
First, a confession: this is the first time I've looked at this repo, and I haven't yet got a locally-running setup, so this is a bit of a shot in the dark right now.

Heres the [trello card](https://trello.com/c/eesRQZ93/706-importing-audio-episodes-bugs-in-atom-workshop).

I've looked for obvious potential problems with the handling of imported data, and added some minor defensive checks where it looked like we may be exposed.

Opening up for comments etc. while I try to find a way to test it...

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,8 +10,14 @@ for both the composer and capi AWS accounts from [janus](https://janus.gutools.c
 
  - Fetch config from S3: `./fetch-config.sh`
  - If you get an error message saying that you requred AWS Signature Version 4, configure your aws cli by running `aws configure set default.s3.signature_version s3v4`
- - Setup the nginx mapping by following the instructions in the
+ - Setup the nginx mapping by following the instructions in the "Install config for an application" section of the 
  [dev-nginx readme](https://github.com/guardian/dev-nginx#install-config-for-an-application).
+ 
+    tl;dr? Try this...
+    * Clone the [dev-nginx](https://github.com/guardian/dev-nginx) private repo to your machine (e.g. ~/code/dev-nginx)
+    * Change directory to that location: `cd ~/code/dev-nginx`
+    * Assuming your copy of atom-workshop is in ~/code/atom-workshop, run `sudo ./setup-app.rb ~/code/atom-workshop/nginx/nginx-mapping.yml`
+    
  - Install Client Side Dependencies with `./scripts/setup.sh`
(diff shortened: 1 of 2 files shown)


regiskuckaertz:
These changes look sensible to me, let's wait until you have had the chance to test though.

JustinPinner:
Ok, I've run this up locally, and have tested it as far as I can take it here. I'm going to have to get it deployed to find out if it has covered all the bases (from the POV of importing an episode's data anyway).

README.md
```diff
@@ -10,8 +10,14 @@ for both the composer and capi AWS accounts from [janus](https://janus.gutools.c
...
  - Fetch config from S3: `./fetch-config.sh`
  - If you get an error message saying that you requred AWS Signature Version 4, configure your aws cli by running `aws configure set default.s3.signature_version s3v4`
- - Setup the nginx mapping by following the instructions in the
+ - Setup the nginx mapping by following the instructions in the "Install config for an application" section of the 
  [dev-nginx readme](https://github.com/guardian/dev-nginx#install-config-for-an-application).
+ 
+    tl;dr? Try this...
+    * Clone the [dev-nginx](https://github.com/guardian/dev-nginx) private repo to your machine (e.g. ~/code/dev-nginx)
+    * Change directory to that location: `cd ~/code/dev-nginx`
+    * Assuming your copy of atom-workshop is in ~/code/atom-workshop, run `sudo ./setup-app.rb ~/code/atom-workshop/nginx/nginx-mapping.yml`
```
akash1810: shouldn't need `sudo` here anymore

JustinPinner: Is that recent? I had to `sudo` that a few days ago or it wouldn't run - had permission denied errors.

akash1810: It's at least a couple of months old - I did it as part of the move to `mkcert` in dev-nginx.

`setup-app.rb` will write a file to `/usr/local/etc/nginx/servers` and then restart nginx. Your user has permission to write to that directory and the [restarting of nginx](https://github.com/guardian/dev-nginx/blob/master/restart-nginx.sh) invokes `sudo` and will there prompt for a password when needed.

If you're having to run `sudo ./setup-app.rb` it would suggest that something has previously been installed as `sudo` and thus has the wrong permissions. That could escalate...

PR #242, opened 2019-05-22
Allow the tag list to scroll across 100 tags

walaura:
Wheeeeeee

Screenshot 2019-05-22 at 15 38 31

### DIFF
--- a/public/styles/components/_forms.scss
+++ b/public/styles/components/_forms.scss
@@ -28,7 +28,8 @@
     background: $color250Grey;
     padding: 10px;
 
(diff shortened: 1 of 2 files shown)


public/styles/components/_forms.scss
```diff
@@ -245,6 +248,9 @@
 
 .form__field__suggestions {
   border-top: 1px solid $color400Grey;
+  border-bottom: 1px solid $color400Grey;
+  max-height: 380px;
```
walaura:
```suggestion
  max-height: 350px;
```

PR #245, opened 2019-05-31
Update queryParams.js

akash1810:
only search for atoms that are editable in the tool


JustinPinner:
Nice

tomrf1:
💯

akash1810:
The atom type filter gets pre-filled, which is a bit gross but we can ship this and address that later.

image

PR #247, opened 2019-08-07
Ensure nvm is installed so that the correct node version is used

amyhughes:
I was tripped up running the setup script, without this change I had to run `source ~/.bashrc` then `nvm use` before the script would work!

Here we set the node version from .`nvmrc` if nvm is installed, and install nvm if not.

### DIFF
--- a/scripts/client-dev.sh
+++ b/scripts/client-dev.sh
@@ -9,6 +9,20 @@ do
     fi
 done
 
+fileExists() {
+  test -e "$1"
+}
+
+if ! fileExists "$NVM_DIR/nvm.sh"; then
+  node_version=`cat .nvmrc`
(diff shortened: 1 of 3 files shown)


amyhughes:
The readme suggests running `setup.sh` (which includes the appropriate steps) before running `start.sh` or `client-dev.sh`. Do you still think we need to add this to these scripts?

akash1810:
Setup is a one off process to get all dependencies installed. Once we've run it, we can do other things (setup different apps that need different node versions, etc). When we're ready to run this app again, I think it would be good if we could run the start script and have it do the right thing, that is, switch to the node version for this app.

(this is really poorly phrased...)

amyhughes:
That makes sense, thanks! I'll add now.

PR #248, opened 2019-10-14
Remove storyquestions

regiskuckaertz:
This atom is deprecated.


regiskuckaertz:
Tested on CODE.

regiskuckaertz:
@mchv we have a much better solution for them now (callouts) with formstack integration.

mchv:
@regiskuckaertz I disagree as story questions were completely different from callouts. It was used to inform editorial about understanding of readers and questions they may have. They are real time integrated in `Ophan` and have found unknown knowledge gap when they have been used by the foreign desk. I reckon they are not used anymore so it makes sense to do not display them in atom workshop.

mchv:
@regiskuckaertz another difference were that answers were predefined, and controlled by Editorial.

regiskuckaertz:
We can achieve all of that with callouts more easily. I use the term "callout" but I should use "campaign" instead.

> They are real time integrated in Ophan

That was ad hoc work, if we integrate targeting we get not only callouts but also real time data for _all_ campaigns in one go.

> answers were predefined, and controlled by Editorial

This is easy to do in formstack: a list of options

Compared to atoms, here's what we _don't_ have to do:

- hack to hide the atom after a few days (such logic had to be duplicated everywhere)
- mailing list management (and the gdpr shenanigans that follow 😭 )
- editing UI
- never having time to do anything journalists were asking for 😆

regiskuckaertz:
To be clear, the idea is brilliant, but the execution wasn't on par with people's expectations, or required too many hacks. If this is prioritised again in the future, we will future-proof it.

PR #259, opened 2019-12-05
CLEAN-UP: Use yarn dev-build instead of hmr 

ajwl:
The hot reload  command that was running `webpack dev server` doesn't work - can we just get rid of webpack-dev-server? 

--> check this tomorrow


ajwl:
remove webpack-dev-server used in old way

PR #260, opened 2019-12-05
CHARTS: Get new chart atom after saving in iframe

ajwl:
Currently the atom doesn't refresh after you close the Charts Editor iframe. 

This makes it look like the change hasn't been saved. Actually the changed atom has been saved to CAPI but just hasn't been updated on the page. 

The fix:
This fires off an "onChartUpdate" method that hits CAPI for the latest version of the Charts Atom whenever the modal (iframe containing the charts tool editor) is closed

### DIFF
--- a/public/js/components/Header/EditHeader.js
+++ b/public/js/components/Header/EditHeader.js
@@ -16,7 +16,7 @@ class EditHeader extends React.Component {
     atom: atomPropType,
     presence: PropTypes.bool,
     saveState: PropTypes.object,
-    formErrors: PropTypes.object,
+    formErrors: PropTypes.array,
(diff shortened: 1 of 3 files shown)


public/js/components/Header/EditHeader.js
```diff
@@ -16,7 +16,7 @@ class EditHeader extends React.Component {
     atom: atomPropType,
     presence: PropTypes.bool,
     saveState: PropTypes.object,
-    formErrors: PropTypes.object,
+    formErrors: PropTypes.array,
```
ajwl: This was causing a type bug so decided to fix it in here..

PR #268, opened 2020-02-13
TAG SEARCH BUG: amend capi search to get better results

ajwl:
The tag search results from CAPI were making it hard to add basic tags like "Football" as results were sorted alphabetically pushing common words very far down the results list. 

This fix copies the CAPI query that we use in Media-Atom-Maker for tags. It appears to add more relevant results first, by looking at "web-title" instead of "internal-name". 

All tags were requested - not just keywords so have removed the "type" check. 

Screenshot 2020-02-13 at 12 44 56
Too many results for Football pushed the keyword "Football" out of reach

### DIFF
--- a/public/js/services/capi.js
+++ b/public/js/services/capi.js
@@ -1,11 +1,9 @@
 import { pandaFetch } from './pandaFetch';
 import { uriEncodeParams, sanitiseQuery } from '../util/uriEncodeParams';
 
-export const searchTags = (searchText, type = null) => {
+export const searchTags = (searchText) => {
   return pandaFetch(
-    `/support/capi/tags?page-size=100&internal-name=${searchText}${
-      type ? `&type=${type}` : ''
-    }`,
+    `/support/capi/tags?web-title=${searchText}&page-size=100`,
     {
       method: 'get',
       credentials: 'same-origin',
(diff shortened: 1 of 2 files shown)


ajwl:
Different calls:

Composer does this on Tag Manager: `/tags?limit=50&query=footb&type=keyword,tone,series,blog,paidContent,campaign`

MAM does this: `/support/previewCapi/tags?page-size=200&type=keyword&web-title=foo`

ajwl:
Just waiting for approval from Celine..

public/js/services/capi.js
```diff
@@ -1,11 +1,9 @@
 import { pandaFetch } from './pandaFetch';
 import { uriEncodeParams, sanitiseQuery } from '../util/uriEncodeParams';
 
-export const searchTags = (searchText, type = null) => {
+export const searchTags = (searchText) => {
```
jennygrahamjones: It looks like the [`FormFieldTagPicker`](https://github.com/guardian/atom-workshop/blob/b9fc1760213034bd282f53ac84ab3ef94e6a7b4c/public/js/components/FormFields/FormFieldTagPicker.js#L47) is still using this function with two params. Changing the signature doesn't seem to have broken anything, but is it worth removing that second parameter for clarity?

ajwl: oh good catch. shoulda checked that

ajwl: updated!

PR #270, opened 2020-03-30
Add new chart types to a description

ajwl:
Now we've added some new chart types in the embedded charts tool - we should call them out in the description


ajwl:
thanks @jennygrahamjones !

PR #301, opened 2021-11-02
fixes the isHttpsUrl regex

SHession:
## What does this change?

This PR fixes an issue which causes validating URLs containing brackets to crash the site.

The validation function `isHttpsUrl` contained a `regex` that would blow up if it detected a string with brackets in the wrong place. [Regex 101](https://regex101.com/) would detect this as **catastrophic backtracking**.

For example this string would cause issues: https://example.co.uk/broken/test(1).mp3

~~To fix this issue I've removed the extra `*` as it seems to be redundant and included brackets in the acceptable character set.~~

I have replaced the regex with a check using the URL class.

## How to test

Are we able to add media atoms where the Track URL contains brackets? We can test this locally or on CODE.

Users are able to successfully add track URLs with brackets.

Working with regex...

### DIFF
--- a/public/js/util/validators.js
+++ b/public/js/util/validators.js
@@ -8,14 +8,21 @@ import FieldError from '../constants/FieldError';
 
 export const isHttpsUrl = value => {
   const stringValue = typeof value === 'string' ? value : '';
-  if (
-    stringValue.match(
-      /^(https:\/\/)?([\da-z\.-]+)\.([a-z\.]{2,6})([\/\w \.-]*)*\/?$/
-    )
-  ) {
+
+  try {
+
+    const url = new URL(stringValue);
+
+    if (url.protocol !== "https:") {
+      const error = new FieldError('not-https', 'Not a HTTPS url');
+      return Promise.resolve(error);
+    }
+
     return Promise.resolve(true);
-  } else {
-    const error = new FieldError('not-https', 'Not a HTTPS url');
+
+  }
+  catch (e) {
+    const error = new FieldError('not-url', 'Not a valid url');
     return Promise.resolve(error);
   }
 };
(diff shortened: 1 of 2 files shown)


akash1810:
I think it would be better to use the [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL) class, specifically the [protocol](https://developer.mozilla.org/en-US/docs/Web/API/URL/protocol) property as it yields simpler code:

```js
new URL("https://uploads.guim.co.uk/broken/test(1).mp3").protocol
> 'https:'
```

You'll need to wrap it in a `try` as `new URL("garbage")` throws, but I think it is still easier to maintain.

SHession:
@akash1810, excellent suggestion. I have that change and added some additional tests.

public/js/util/validators.js
```diff
@@ -8,14 +8,22 @@ import FieldError from '../constants/FieldError';
...
-    const error = new FieldError('not-https', 'Not a HTTPS url');
+
+  try {
+
+    const url = new URL(stringValue);
+
+    if(url.protocol === "https:"){
+      return Promise.resolve(true);
+    }
+    else{
```
akash1810, line 19: minor: this `else` isn't needed with the early return.

public/js/util/validators.spec.js
```diff
@@ -1,19 +1,42 @@
...
-    .then(res => {
-       expect(res).toBe(true);
-    });
-});
+describe("isHttpsUrl function", () => {
+
+  it('Should validate a simple URL as HTTPS URL', () => {
+    let url = 'https://example.com';
+
+    return isHttpsUrl(url)
```
akash1810, line 9: minor: `async`/`await` would make these tests slightly easier to read, but not sure what pattern is employed across the repository

akash1810 (approved):
Excellent! Additional tests look great.

A couple of non-blocking minor comments - feel free to ignore.

public/js/util/validators.spec.js
SHession: Agreed, although, it does look this is consistent with the rest of the project. I have removed the `return`s are they are unnecessary

PR #302, opened 2021-11-03
Adds ci script

SHession:
## What does this change?

This PR changes the CI/CD process to use a script instead of Team City steps.

This matches our best practice for our applications and will allow us to experiment with changes to our CI without affecting main. 

## How to test

Does the project build and deploy successfully? (We have updated TC to use the script and deployed to CODE)

## Images

image

### DIFF
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+
+set -e
+
+setupNvm() {
+    export NVM_DIR="$HOME/.nvm"
+    [[ -s "$NVM_DIR/nvm.sh" ]] && . "$NVM_DIR/nvm.sh"  # This loads nvm
+
+    nvm install
+    nvm use
+}
+
+globalJsDependencies() {
+    npm install -g yarn
+}
+
+yarnSetup() {
+    yarn install --force
+    yarn lint
+    yarn test
+    yarn build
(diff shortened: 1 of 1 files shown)


scripts/ci.sh
```diff
@@ -0,0 +1,35 @@
...
+    nvm install
+    nvm use
+}
+
+globalJsDependencies() {
+    npm install -g yarn
+}
+
+yarnSetup() {
+    yarn install --force
```
akash1810, line 18: This doesn't yield a deterministic build as the lockfile get's completely ignored - `yarn install --frozen-lockfile` is preferred. From https://classic.yarnpkg.com/en/docs/cli/install/:

> If you need reproducible dependencies, which is usually the case with the continuous integration systems, you should pass `--frozen-lockfile` flag.

akash1810 (commented):
Could I tempt you towards [GitHub Actions](https://docs.google.com/document/d/1Xz-gzptICIza0oio0Kza3LWg2a6n_015o0MPu0LGfuU/edit)? More than happy to pair!

scripts/ci.sh
SHession: Good advice, thanks @akash1810!

SHession:
This PR is no longer needed after the switch to GitHub actions 🎉

PR #303, opened 2021-11-03
Adds GitHub action CI workflow

SHession:
## What does this change?

This PR attempts to move our CI away from TeamCity and into GitHub actions using the [latest tools from DevX](https://docs.google.com/document/d/1Xz-gzptICIza0oio0Kza3LWg2a6n_015o0MPu0LGfuU/edit#).

**NB** The action will run for the PR but will not attempt to upload until we add `riffRaffUpload` to the sbt step.

## How to test

We will need to pause TC build and set the build number, we should then be able to test the process end-to-end using CODE.

We move away from TeamCity builds and gain all the benefits of GitHub actions!

This could be disruptive for Atom Workshop builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,8 +7,6 @@ on:
   push:
     branches:
       - master
-env:
-  FINAL_TEAMCITY_BUILD: 1309
 jobs:
   ci:
     runs-on: ubuntu-latest
@@ -46,5 +44,6 @@ jobs:
     - name: CI for sbt-riffraff-artifact
       shell: bash
       run: |
-        export GITHUB_RUN_NUMBER=$(( $GITHUB_RUN_NUMBER + env.FINAL_TEAMCITY_BUILD ))
-        sbt clean compile
+        FINAL_TEAMCITY_BUILD=1319
+        export GITHUB_RUN_NUMBER=$(( $GITHUB_RUN_NUMBER + $FINAL_TEAMCITY_BUILD ))
+        sbt clean compile riffRaffUpload
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,13 @@ jobs:
       with:
         java-version: 8
         distribution: adopt
+    - name: Cache SBT
+      uses: actions/cache@v2
+      with:
+        path: |
+          ~/.ivy2/cache
+          ~/.sbt
+        key: ${{ runner.os }}-sbt-${{ hashFiles('**/build.sbt') }}
     - name: CI yarn
       shell: bash
       run: | 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
   workflow_dispatch:
   push:
     branches:
-      - $default-branch
+      - master
 env:
   FINAL_TEAMCITY_BUILD: 1309
 jobs:
@@ -23,11 +23,7 @@ jobs:
     - uses: guardian/actions-assume-aws-role@v1
       with:
         awsRoleToAssume: ${{ secrets.GU_RIFF_RAFF_ROLE_ARN }}
-    - name: Get node version
-      run: echo NODE_VERSION=$(cat .nvmrc) >> $GITHUB_ENV
-    - uses: actions/setup-node@v2
-      with:
-          node-version: ${{ env.NODE_VERSION }}
+    - uses: guardian/actions-setup-node@v2.4.1
     - name: Set up JDK
       uses: actions/setup-java@v2
       with:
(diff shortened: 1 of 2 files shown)


SHession:
@akash1810, it looks like from a cold box the build times are quite similar between TC and GHA but they can be significantly quicker in TeamCity if it's warm. Any thoughts:
image

.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
...
+    - name: Checkout branch
+      uses: actions/checkout@v2
+    - uses: guardian/actions-assume-aws-role@v1
+      with:
+        awsRoleToAssume: ${{ secrets.GU_RIFF_RAFF_ROLE_ARN }}
+    - name: Get node version
+      run: echo NODE_VERSION=$(cat .nvmrc) >> $GITHUB_ENV
+    - uses: actions/setup-node@v2
+      with:
+          node-version: ${{ env.NODE_VERSION }}
```
akash1810, line 30: This can be replaced with https://github.com/guardian/actions-setup-node:

```
- uses: guardian/actions-setup-node@v2.4.1
```

`guardian/actions-setup-node` will automatically choose the correct node version by inspecting the `.nvmrc` file and others. It's a fork of `actions/setup-node` (see https://github.com/actions/setup-node/pull/239).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
...
+      shell: bash
+      run: | 
+        yarn install --force --frozen-lockfile
+        yarn lint
+        yarn test
+        yarn build
+    - name: CI for sbt-riffraff-artifact
+      shell: bash
+      run: |
+        export GITHUB_RUN_NUMBER=$(( $GITHUB_RUN_NUMBER + $FINAL_TEAMCITY_BUILD ))
```
akash1810, line 46: You're missing a line here I think?

```bash
FINAL_TEAMCITY_BUILD=???
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
+name: CI for sbt-riffraff-artifact
+
+on:
+  pull_request:
+  # Manual invocation.
+  workflow_dispatch:
+  push:
+    branches:
+      - $default-branch
```
akash1810, line 9: This should be `main` here, `$default-branch` works within the `.github` repo as some magic templating happens there ✨ .

akash1810:
> @akash1810, it looks like from a cold box the build times are quite similar between TC and GHA but they can be significantly quicker in TeamCity if it's warm. Any thoughts: image

An initial glance suggests a lot of the time in GHA is spent downloading dependencies. Caching these should help:
  - Node caching https://github.com/actions/setup-node#caching-packages-dependencies
  - SBT cache https://github.com/actions/setup-java#caching-packages-dependencies

I've not actually done SBT caching yet, so intrigued to know how to!

.github/workflows/ci.yml
```diff
@@ -0,0 +1,47 @@
...
+      run: | 
+        yarn install --force --frozen-lockfile
+        yarn lint
+        yarn test
+        yarn build
+    - name: CI for sbt-riffraff-artifact
+      shell: bash
+      run: |
+        export GITHUB_RUN_NUMBER=$(( $GITHUB_RUN_NUMBER + $FINAL_TEAMCITY_BUILD ))
+        sbt clean compile
```
akash1810, line 47: You'll want `riffRaffUpload` on the end too.

.github/workflows/ci.yml
SHession: Unfortunately we still have master on this repo. We have card to change it though. I'll set it to master for now.

.github/workflows/ci.yml
SHession: I excluded this to avoid creating builds until I was ready to disable TC. I will add it back in shortly.

.github/workflows/ci.yml
SHession: I've added at the top as an ENV variable. Might be better to have it as part of that action, I don't have a strong opinion.

SHession:
@akash1810, thats for the review, really helpful. I've got it working now and have successfully deployed to CODE. Happy to discuss any other suggestion for improvements but possibly we could iterate on them in another PR?

akash1810 (approved):
YEAH! Great work!

img

akash1810:
> I've added SBT caching which seems to have improved the build times somewhat. I haven't yet added yarn caching, might improve build speeds a bit more.

I think `yarn` caching is provided [for free](https://github.com/actions/setup-node#caching-packages-dependencies) if when the lockfile is on the root of the repo, which it is.

PR #304, opened 2021-11-04
Adds fluentbit conf

SHession:
## What does this change?

This PR aims to ship logs via [Fluent Bit](https://docs.fluentbit.io/manual/pipeline/outputs/kinesis) instead of using the standard [Logback Kinesis appender](https://github.com/guardian/kinesis-logback-appender). This was related to the general tools outage, more information can be found in the [incident document](https://docs.google.com/document/d/17jGwlvc5F-SOCSZ90lNl5ZRpJMvvaJ_UXtuUyY4ZM1s/edit#heading=h.98xvxa3bjhw).

This work was completed as part of a spike, the main objective of which is improve our logging by being more robust in the face of issues with the Kinesis stream. This change achieves this by removing writing to the Kinesis stream as an application concern, and instead relies on Fluent Bit running on the instances. Fluent Bit becomes responsible for handling our logs and our application is not blocked by these concerns.

The Fluent Bit conf looks at the journald output of the atom-workshop.service which provides us our usual application logs, in addition to other logs which weren't previously included. Instead of using the Kinesis appender, we now use STDOUT to output our logs in a way which will surface in journald. This is also wrapped in an `AsyncAppender` which has certain [useful characteristics](https://stackoverflow.com/questions/47372092/when-not-to-use-asyncappender-in-logback-by-default/47409324#47409324).

Existing issue:
Currently this initial Play logs are output to journald without being well formatted, this means we see logs in ELK which are split across line. It would be better if these could be grouped together and formatted properly:
image
 

## How to test

Deploying to CODE in conjunction with the necessary [CFN changes](https://github.com/guardian/editorial-tools-platform/pull/564). Do the [logs show up in ELK](https://logs.gutools.co.uk/goto/ec7a6d8c65121a9fbd2cac29f2430d1a)?

Our logs no longer block the running application, they are therefore more reliable. More logs are also included as a result, which may be useful in debugging.

Can we rely on this system enough to ensure we always receive our logs in the way we expect?

## References

Janus-app PR: https://github.com/guardian/janus-app/pull/21
Janus PR: https://github.com/guardian/janus/pull/2774


fluentbit/td-agent-bit.conf
```diff
@@ -0,0 +1,56 @@
...
+  Key_Name message
+  Parser systemd_json
+
+# https://docs.fluentbit.io/manual/pipeline/filters/modify
+# Add app identity fields
+[FILTER]
+  Name modify
+  Match *
+  Add app atom-workshop
+  Add stage STACK_STAGE
```
akash1810, line 38: Is it worth adding a comment to say this value is replaced within the launch configuration?

SHession: Good suggestion, I've done this.

PR #306, opened 2021-11-16
Passes logging timestamp transparently

SHession:
## What does this change?

This PR aims to fix issues with the timestamp in the ELK logs:  

before:
image

after: 
image

The change fixes this issue by no longer parsing the time from the log JSON, and then parsing it back out as a key. Instead we simply pass the keys straight through to Fluent Bit.

## How to test

Deploy to CODE via riff-raff, observe the logs. 
The timestamp in ELK should match the JSON output on the box. This can be retrieved by SSHing onto the box and running: `journalctl -u atom-workshop.service`

## Note 

We aren't currently parsing the time format from non-JSON log-lines, meaning we can see the keyed timestamp doesn't match the message. We could fix this by creating a custom parser for the format, or changing how they're encoded. As these we're logs we didn't previously include, I'm not treating this issue as a priority: 
image

### DIFF
--- a/fluentbit/parsers.conf
+++ b/fluentbit/parsers.conf
@@ -2,5 +2,3 @@
 [PARSER]
   Name systemd_json
   Format json
-  Time_Key @timestamp
-  Time_Format %Y-%m-%dT%H:%M:%S.%L%z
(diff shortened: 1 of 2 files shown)


rhystmills (approved):
Looks good to me!

PR #312, opened 2022-04-01
Play 2.6 upgrade

davidfurey:
## What does this change?

Upgrades to Play 2.6.  Requires an upgrade to atom-maker (https://github.com/guardian/atom-maker/pull/62) before it can be deployed.

Out of the box Play 2.6 only accepts requests with the correct `Host` header, which is a bit of a pain and provides little security benefit, so it has been disabled.

Play 2.6 also has a very strict Content-Security-Policy, which I've replaced with something more permissive.

Tested locally and on CODE.

### DIFF
--- a/app/views/index.scala.html
+++ b/app/views/index.scala.html
@@ -3,7 +3,7 @@
         jsLocation: String,
         presenceJsLocation: Option[String],
         clientConfigJson: String,
-        csrf: Option[String] = None
+        csrf: String
 )
 <!DOCTYPE html>
 <html lang="en">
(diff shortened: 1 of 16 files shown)


andrew-nowak:
I tried testing on CODE but the deploy is failing and the new instances aren't booting. There's also no helpful logs

davidfurey:
@andrew-nowak the application wasn't running so no logs. I've added a commit to fix it.

andrew-nowak (approved):

app/views/index.scala.html
```diff
@@ -1,11 +1,25 @@
-@(title: String, jsLocation: String, presenceJsLocation: Option[String], clientConfigJson: String)
+@(
+        title: String,
+        jsLocation: String,
+        presenceJsLocation: Option[String],
+        clientConfigJson: String,
+        csrf: Option[String] = None
```
andrew-nowak, line 6: should this be optional? I can see only one usage and it's hardcoded as Some

public/js/services/AtomsApi.js
```diff
@@ -21,7 +21,8 @@ export default {
         credentials: 'same-origin',
         body: JSON.stringify(atomInfo),
         headers: {
-          'Content-Type': 'application/json'
+          'Content-Type': 'application/json',
+          'Csrf-Token': window.guardian.csrf.token
```
andrew-nowak, line 25: These would throw if window.guardian is undefined -- though possibly not an issue since csrf will always be defined?

andrew-nowak (approved):

app/views/index.scala.html
davidfurey: Good  point. This was copy+pasted from a project which presumably need the option. I'll change it  here.

public/js/services/AtomsApi.js
davidfurey: Indeed. I got half way through adding the multiple layers of `undefined` checks, but decided that it will always be defined, and if it isn't defined the http call would fail anyway.

PR #316, opened 2022-06-06
Double CAPI tags query to 200

andrew-nowak:
## What does this change?

Searching for some tags can include a lot of responses... eg. search for France returns 120 responses, and France (topic) is not in the top 100. As you can't get more specific, this means that you can no longer select France as a topic for an atom. This PR doubles the limit to 200 to allow picking France until we can have a better, longer term solution (sorting, paginating, etc.)

## How to test

On CODE, france only returns 90 results, so France can always be found. Try searching "Fra" and looking for "Fran Roots (contributor)" - they should be around 110th result, so visible with this change.

Editorial are unblocked

This problem will happen again as more tags are added. Composer manages to sort the results in a sensible fashion ("France" is at the top of the results, so can never drop off the bottom of any paging). Can atom-workshop do the same?

### DIFF
--- a/public/js/services/capi.js
+++ b/public/js/services/capi.js
@@ -3,7 +3,7 @@ import { uriEncodeParams, sanitiseQuery } from '../util/uriEncodeParams';
 
 export const searchTags = (searchText) => {
   return pandaFetch(
-    `/support/capi/tags?web-title=${searchText}&page-size=100`,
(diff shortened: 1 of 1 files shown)


Fweddi:
Similar issue addressed in Composer in https://github.com/guardian/flexible-content/pull/3773 - this will be okay for now, but as you say, the issue will continue to grow.

Fweddi (approved):

PR #328, opened 2022-10-10
Upgrade to sbt 1.6.2

davidfurey:
## What does this change?

Fairly minimal changes to upgrade to SBT 1.6.2 in order to resolve jnotify issue on M1 macs.  

- Switched sonatype resolver to be https, because SBT wasn't happy with http.
- Bumped sbt-digest because 1.1.1 isn't published for SBT 1.x
- Bumping the patch release of Play was probably unnecessary but equally not a bad thing to do anyway.

NB- I haven't fixed any deprecation warnings (sorry).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -26,7 +26,7 @@ libraryDependencies ++= Seq(
 )
 
 resolvers ++= Seq(
-  "Sonatype OSS Snapshots" at "http://oss.sonatype.org/content/repositories/snapshots/",
+  "Sonatype OSS Snapshots" at "https://oss.sonatype.org/content/repositories/snapshots/",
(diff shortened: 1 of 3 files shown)


ParisaTork:
CI build is failing but PR LGTM 👍

rhystmills:
I had no trouble running locally so interesting that the Play compile seems to have failed in CI

davidfurey:
I seems that Play 2.6 isn't compatible with newer versions of SBT.  I haven't dug into why it works locally but CI packaging fails.  My preference would be to focus on upgrading this to Play 2.8, but we'll need to upgrade to at least Scala 2.12 first, and migrate from configuration-magic to simple-configuration

PR #331, opened 2022-11-04
Permissions and Config upgrades

rebecca-thompson:
## What does this change?
This PR:

- upgrades the `editorial-permissions-client` and `atom-*` libraries in preparation of bumping Scala to 2.12
- migrates from `configuration-magic-core` to `simple-configuration-ssm`, this also supports bumping Scala versions and is required as this library has been archived and is no longer maintained
- does a general tidy up of the config in this project:
    -  it moves the loading of config to `App.scala`. This fixes a bug where the `play.crypto.secret` was not being updated to a secure value
    - remove unused config from `Config.scala` and organise remaining config into logical groups

Outside of this PR, this change involves moving our config in AWS from DynamoDB to Parameter Store.

## How to test
Deploy the branch to CODE

- [x] do permissions still work? (currently the only permission is the ability to delete atoms)
- [x] does pan-domain auth still behave as expected?
- [x] can you create new atoms and see existing atoms?

Note that if you're running this project locally, you will need to re-run the `fetch-config.sh` script to pull in the new DEV config from S3. The config has been tidied up, so the file itself has changed as well as the location from `~/.configuration-magic/atom-workshop.conf` to `~/.gu/atom-workshop.conf` 

We are one step closer to upgrading Scala, SBT and Play versions!

This is a medium risk PR as it involves migrating config. It might be worth review the config in AWS alongside this code review.

### DIFF
--- a/fetch-config.sh
+++ b/fetch-config.sh
@@ -2,5 +2,5 @@
 
 region="eu-west-1"
 
-mkdir -p ~/.configuration-magic/
-aws s3 cp s3://guconf-flexible/atom-workshop/atom-workshop.conf ~/.configuration-magic/atom-workshop.conf --profile composer --region $region
+mkdir -p ~/.gu/
+aws s3 cp s3://guconf-flexible/atom-workshop/atom-workshop.conf ~/.gu/atom-workshop.conf --profile composer --region $region


jonathonherbert (approved):
Run through the scenarios you give in the todo list, and all work as expected 👍 

I've also been through the configuration param groups and tested functionality that related to each, e.g. updating CAPI (by using an atom created by this build in Composer).

Can see the related params in SSM, all look good and this is a great improvement 👌 

One thing to note – I cannot take down or delete atoms. I thought this was a problem with this PR, but I think it's a broader problem, possibly introduced by #312 – I can replicate locally on main and in PROD 😱 I'll follow that up.

jonathonherbert:
Problem with deletion mentioned above addressed in #332.

PR #341, opened 2023-05-15
Improve error handling when an atom cannot be updated due to a version conflict

jacobwinch:
## What does this change?

We (@guardian/devx-reliability) noticed that this application is serving a (relatively) high percentage of 5XX errors[^1]. Most of these errors seem to occur due to a server side safety check which prevents users from accidentally overwriting changes that have been stored in the database since the client last fetched the atom. In my opinion, this should be treated as a client side error and not a server side error (i.e. the server is working as desired), so this PR modifies the code to return a [409](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409) instead of a 500 in this scenario.

Although there could be a number of explanations for this client side error, I suspect that many of these errors could just be due to multiple users (or the same user with multiple tabs open) editing an atom simultaneously. Analysis of the [application logs](https://logs.gutools.co.uk/s/editorial-tools/goto/44d76920-f304-11ed-84b5-9796daab3a37) suggests that the same users are often repeatedly attempting to update an atom in this state. I suspect these repeated attempts are partly due to the ambiguous error message that is currently displayed, so this PR also modifies the error message to try to explain the problem to users more clearly. That said, I don't have any UX expertise so any feedback on the new error message would be appreciated! 

## How to test

I have deployed this change to `CODE` and attempted to edit the same atom in multiple tabs. When doing this the new error message is displayed as expected (see screenshots below).

* The Availability SLI for this service (as seen [here](https://metrics.gutools.co.uk/d/P0AtBxJVk/sli-dashboard-summary-view?orgId=1)) should improve as we will be serving fewer 5XX errors
* Hopefully we will also see fewer messages related to version conflicts in the logs (if the new message works as intended users won't continue attempting to update the atom, as they'll understand that it is not going to work until they refresh/close other tabs)

I don't think there are any particular risks associated with this PR.

## Images

**Before:**

image

**After:**

image

## Accessibility

As I have only changed error message copy I have not conducted any testing related to accessibility.

[^1]: This means that it is flagged as a problem in our [Availability SLI dashboard](https://metrics.gutools.co.uk/d/P0AtBxJVk/sli-dashboard-summary-view?orgId=1).

### DIFF
--- a/app/db/AtomWorkshopDB.scala
+++ b/app/db/AtomWorkshopDB.scala
@@ -1,19 +1,22 @@
 package db
 
 import cats.syntax.either._
-import com.gu.atom.data.{DataStoreResultUtil, DynamoDataStore, IDNotFound}
+import com.gu.atom.data.{DataStoreResultUtil, DynamoDataStore, IDNotFound, VersionConflictError}
 import com.gu.contentatom.thrift.{Atom, AtomType}
 import com.gu.pandomainauth.model.User
-import models.{AtomAPIError, AtomWorkshopDynamoDatastoreError, UnknownAtomError}
+import models.{AtomAPIError, AtomWorkshopDynamoConflictError, AtomWorkshopDynamoDatastoreError, UnknownAtomError}
 import play.api.Logging
 import util.AtomLogic._
 
-trait AtomWorkshopDBAPI {
+trait AtomWorkshopDBAPI extends Logging {
 
   def transformAtomLibResult[T](atomType: AtomType, id: String, result: DataStoreResultUtil.DataStoreResult[T]): Either[AtomAPIError, T] = result match {
     case Left(IDNotFound) => Left(UnknownAtomError(atomType, id))
+    case Left(VersionConflictError(_)) => Left(AtomWorkshopDynamoConflictError)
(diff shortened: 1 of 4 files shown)


app/db/AtomWorkshopDB.scala
```diff
@@ -1,19 +1,22 @@
...
-trait AtomWorkshopDBAPI {
+trait AtomWorkshopDBAPI extends Logging {
 
   def transformAtomLibResult[T](atomType: AtomType, id: String, result: DataStoreResultUtil.DataStoreResult[T]): Either[AtomAPIError, T] = result match {
     case Left(IDNotFound) => Left(UnknownAtomError(atomType, id))
+    case Left(VersionConflictError(_)) => Left(AtomWorkshopDynamoConflictError)
     case Left(e) => Left(AtomWorkshopDynamoDatastoreError(e.msg))
-    case Right(r) => Right(r)
+    case Right(r) =>
+      logger.info(s"Successfully updated atom of type ${atomType.name} with id $id")
```
jacobwinch, line 18: This logging has been moved so it only gets logged if the atom was actually updated

davidfurey (approved):

PR #342, opened 2023-10-20
Bump node from 12 > 16 and bump incompatible packages (including webpack 2 > 3)

rhystmills:
## What does this change?

Node 12 has been out of maintenance for a while - this PR bumps it to v16. This is [just about out of LTS itself](https://nodejs.dev/en/about/releases/) but is an improvement on the current situation. 

The PR also bumps Webpack to v3 in order to be compatible with the new node version. I still get some error output on installing dependencies but the app seems to run fine afterwards. I'll be following up with a PR bumping to Webpack v5, which solved the unhappy output (separate PR because v3 to v5 Webpack bump is generally quite complicated).

On bumping to Node 18+ - I'm getting build warnings that seem to be related to webpack v3. Once Webpack is on v5 we may be able to bump Node further.

## How to test

Run this branch according to the instructions in the readme. Does it build and run locally?


rhystmills:
Closed in favour of https://github.com/guardian/atom-workshop/pull/343

PR #345, opened 2023-11-02
Remove deprecated recipe atom

rowannekabalan:
## What does this change?

This removes the recipe atom views and wiring as it is being replaced with a new recipe element. 

## How to test

Tested in CODE. 

Recipe atoms should not appear in the UI nor the options to create or edit them.


JustinPinner (commented):
I believe this is safe to do once any existing data has been removed and the recipe-less models are published. 
It also looks like there should be corresponding `project/Dependencies.scala` changes. There are a couple of other libs mentioned in `Dependencies.scala` that I don't recognise (`atom-manager-play` and `atom-publisher-lib`), and I assume the new recipe-less `content-atom` models must be in an updated version of `content-api-client` for this too.

tonytw1 (approved):

PR #347, opened 2023-11-24
Remove unused dependency kinesis-logback-appender

davidfurey:
This library is not used and is deprecated. Removing this dependency revealed that aws-java-sdk-sts and aws-java-sdk-inesis are required but only included via kinesis-logback-appender

At the same time, I've removed a key that SBT warned was unused.


simonbyford (approved):
👍

PR #352, opened 2024-03-12
Play 3.0 upgrade

davidfurey:
## What does this change?

Upgrades from Play 2.9 to 3.0. This is a fairly small change since the only difference between these versions is Akka vs Pekko


dblatcher:
deployed to code, was able to create and publish an atom

dblatcher (approved):

davidfurey:
Deployed to prod, app still seems to work

PR #349, opened 2024-01-18
Upgrade to Play v2.9, Scala 2.13, Java 11

rtyley:
These upgrades were prompted by https://github.com/guardian/atom-maker/pull/94, but they're all probably a good idea anyway:

* **Play:** 2.8 → 2.9
* **Scala:** 2.12 → 2.13
* **Java:** 8 → 11

Note that this PR is using a preview release of https://github.com/guardian/atom-maker/pull/94, which should be merged before this PR.

As of commit d39b421d75c99259a2f326af2c1551415dedf4fd, this PR has [deployed](https://riffraff.gutools.co.uk/deployment/view/958da917-8ef4-4d38-8bc7-7487de2d5b97) successfully to [CODE](https://atomworkshop.code.dev-gutools.co.uk/atoms/guide/23987a06-0cf8-4c2e-b41a-4e843b34523c/edit):

image

### No `editorial-tools-platform` updates required

We've reviewed https://github.com/guardian/editorial-tools-platform/blob/main/cloudformation/composer-account/atom-workshop/atom-workshop.yml to check that we don't need to update a Java version there (ie we haven't forgotten it like we did with https://github.com/guardian/media-atom-maker/pull/1149 ).

PR #353, opened 2024-04-26
Remove references to Scribe

rhystmills:
## What does this change?

This PR replaces our Scribe editor with ProseMirror, via [@guardian/prosemirror-editor](https://github.com/guardian/prosemirror-editor). [Scribe](https://github.com/guardian/scribe) is deprecated and contains vulnerabilities, so we want to replace it everywhere it exists. ProseMirror is the rich text editor we use in other tools (most notably flexible-content) so it makes sense to use it here too. This work has already been done for [tagmanager](https://github.com/guardian/tagmanager/pull/500/files) and there's a PR open in [targeting](https://github.com/guardian/targeting/pull/186).

`atom-workshop` has historically used `<b>` and `<i>` tags rather than the more semantically correct `<strong>` and `<em>` tags - I've preserved these (via `transformToLegacyMarkup`) for the sake of maintaining styles in consuming platforms.

The UX is slightly improved on the previous editor, introducing hover and disabled states for the buttons.

| Before | After |
| --- | --- |
|  |  |

## What else does this change?
I found the setup script was failing on a step where `brew` tries to install `openjdk8`. As it happens this repo [now uses Java 11 ](https://github.com/guardian/atom-workshop/pull/349) and advises using a package manage to install it, so I've removed that step.

(I also added a mention to the readme that other node package managers are available.)


public/js/components/FormFields/FormFieldScribeEditor.js
simonbyford, line 1: Should this be renamed `FormFieldRichTextEditor.js`?

simonbyford (approved):
I don't know anything about Scribe so may have missed some subtleties in the migration but the code looks good and everything appears to work okay. I tried editing/creating atoms and making sure the rich text fields are persisted correctly, all looks good.

Great work on tackling the last high-priority vulnerability! 🦾

public/js/components/FormFields/FormFieldScribeEditor.js
rhystmills: It absolutely should.

PR #354, opened 2024-06-10
access permissions logging

andrew-nowak:
Requires https://github.com/guardian/permissions/pull/195

## What does this change?

Perform logging of user's access level, both with the delete permission (currently the only other permission respected by Atom-workshop) and solely by the access permission.

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

PR #358, opened 2024-07-08
enforce atom_workshop_access permissions

andrew-nowak:
Does what it says in the title.

Follow up to #354 

atom_workshop_access has been added to the content_* permissions groups, and all known users are members of one of those groups, so this shouldn't impact any users.


bryophyta (approved):
+1

PR #408, opened 2026-05-29
AWS SDK v1 -> v2 migration

andrew-nowak:
## What does this change?

Performs the AWS SDK migration from v1 to v2. v1 remains on the classpath as it is depended on by atom-maker libs as this version includes support for both v1 and v2. A future version will drop the support for v1 and it will be removed from the classpath at that point.

## How has this change been tested?

On CODE, creating a new atom and adding it to a Composer article.

Atom-maker lib can drop support for the v1 SDK.

## Accessibility

-   [ ] [Tested with screen reader](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#screen-reader)
-   [ ] [Navigable with keyboard](https://github.com/guardian/accessibility/blob/main/people-and-technology/02-physical.md#keyboard)
-   [ ] [Colour contrast passed](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#contrast)
-   [ ] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

### DIFF
--- a/app/AppComponents.scala
+++ b/app/AppComponents.scala
@@ -5,10 +5,10 @@ import config.{AWS, Config}
 import controllers.{AssetsComponents, ExplainerReindexController, PanDomainAuthActions}
 import db.{AtomDataStores, AtomWorkshopDB, ExplainerDB}
 import play.api.ApplicationLoader.Context
-import play.api.{BuiltInComponentsFromContext, Configuration}
 import play.api.libs.ws.WSClient
 import play.api.libs.ws.ahc.AhcWSComponents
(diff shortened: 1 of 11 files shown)


JustinPinner (approved):
Looks sensible to me. Happy to wait for a second opinion here but will approve on the grounds that this is an internal tool, and thus we can recover quickly in the unlikely event of unexpected upsets (also assuming there are no irreversible side effects to Dynamo or S3 as a result of upgrading the libraries) 👍

PR #409, opened 2026-06-01
Upgrade atom-maker lib to new v12 with updated reindexing logic

andrew-nowak:
## What does this change?

Updates atom-workshop to v12. Implement the updated atom reindexing classes (https://github.com/guardian/atom-maker/pull/146) which reindex atoms in batches, so we no longer need to fit the entire atom database into memory all at once.

## How has this change been tested?

Tested on CODE 

With a successful reindex of atoms.


JustinPinner (approved):
Looks great to me! 👍

